PR #1, opened 2015-04-22
Little Correction in makefile.gcc to prevent warnings after building and...

grauehaare:
... correct Files/Paths


frozenblit:
Thanks! Good catch :)

PR #3, opened 2015-08-30
Some new features

gtozzi:
Hi,

here are 3 commits I think should be merged into the POL, in order of importance:
1. I have added a new "unused" keyword to the escript language. It can be used to instruct the compiler on how to handle unused variables in program and function definitions. See commit message for more details.
2. For testing 1, I have used the testsuite, created a new test case and updated the testsuite script meanwhile to comply with latest python releases
3. I've found a new parameter in the gump commands and added it to the documentation

It's been a very long time since I've coded C++, that's why I've double-checked and tested everything. Hope it is ok. If not, please let me know what's wrong and I will try to fix it.


gtozzi:
Changelog entry added ;)

ThomasVolkert:
Thank you.

PR #9, opened 2016-01-23
Added item.house property, implemented comparison for object of different types, test cases

gtozzi:
- Added item.house property (found bugs corrected in subsequent commits)
- Refactored comparison operators in BObjectImp (and childs) for better coherency when comparing objects of different types (and, imho, also cleaned the code)
- Many improvements in test suite and documentation (I could send these directly to the official branch, but I preferred to write them directly on the updated versions of the sources)


pol-core/pol/multi/house.cpp
```diff
@@ -419,7 +448,19 @@ namespace Pol {
...
+            UHouse* contHouse = item->house();
+            if( contHouse == nullptr ) {
+              os << "This is probably a core bug. Please report it on the forums.";
+            } else {
+              os << "This item is already part of house " << contHouse->serial << "." << std::endl;
+              os << "Allowing an item to be a component in two different houses was a bug," << std::endl;
+              os << "please also fix your save data.";
+            }
+            throw std::runtime_error(os.str());
+          }
```
turleypol: Instead of std::ostringstream fmt::Writer would be better. In addition the serials are currently not shown in hex. ".. 0x" << fmt::hexu(serial) would do it.

gtozzi:
@turleypol Thank you, I've applied your changes.

Bugs fixed, done more testing, updating title and description to make this a real pull request.

PR #52, opened 2018-04-10
Added Split String feature over String class

Daviex:
I needed this, and can't found anything into the docs so I added it myself into the core.
Now you are able to do like:

var x := "Hello World";
x.split(" ");

Output will be an array containing 2 elements:
{ "Hello", "World" }

Probably can be refactored a little. Also, tell me if it needs the implementation to use within the core to be able to add the pull.


Admin-Yukiko:
Look in the POL documentation in the Basic module for SplitWords. That will do what you want. Plus you can specify other delimiters besides spaces.

Daviex:
Nice. I'm an idiot 😅 I searched not really well, I was sure to find such a thing into String class, never thought of Basic module 🤔

Well, was a good experience anyway to do such a thing 👍

Daviex:
Nope, I will enter next time I'm on my PC! Anyway, I think I will look into the SplitWords anyway to add as a SubMethod of the String Class, it would be really better to use, add the Basic Module just for that I think is useless when the String class could contain it without problem!

PR #60, opened 2018-07-12
WIP Handle smooth sailing with mouse piloting

KevinEady:
* Set the boat's process to the one created via `start_script("boat/misc")`
* Boat movement packets are events sent to the boat's process
* Set a boat's pilot via `boat.pilot(charRef)` and clear it via `boat.pilot()`. Setting the pilot will handle the creation of the new boat mount piece 
* Create a new `boat.pilot` read-only member that points to the active pilot
* Various handling of the boat mount piece item.

The boat script will be responsible for handling the movement and turning of boats via receiving the the above-mentioned boat move event, structured as follows:

```
{
  "type": EVID_BOAT_MOVEMENT 
  "source": char ref for the mobile who sent the packet
  "speed": 0 for stop, 1 for one-tile, 2 for normal
  "direction": absolute direction
}
```


KevinEady:
I have a working prototype based off the Distro's boat.src.

If the above API looks good, I will add the doc updates etc prior to requesting a final merge into master.

pol-core/pol/extobj.cpp
```diff
@@ -47,6 +47,8 @@ void read_extobj()
   settingsManager.extobj.sails = elem.remove_unsigned( "Sails", 0x1F019 );
   settingsManager.extobj.storage = elem.remove_unsigned( "Storage", 0x1F01A );
   settingsManager.extobj.weaponslot = elem.remove_unsigned( "Weaponslot", 0x1F01B );
+  settingsManager.extobj.boatmount = elem.remove_unsigned( "Boatmount", 0x1F01C );
```
turleypol: Add it also to our example extobj.cfg

pol-core/pol/multi/boat.cpp
```diff
@@ -1606,9 +1609,11 @@ void UBoat::readProperties( Clib::ConfigElem& elem )
   rescan_components();
 
   regself();  // do this after our x,y are known.
-  // consider throwing if starting position isn't passable.
+              // consider throwing if starting position isn't passable.
 
-  Core::start_script( "misc/boat", make_boatref( this ) );
+  this->process(
+      Core::start_script( Core::ScriptDef( "misc/boat", nullptr ), make_boatref( this ) ) );
+  this->process()->attached_item_ = this;
```
turleypol: start_script can fail and would then return a nullptr, so this line will crash. check is needed

PR #64, opened 2018-07-27
List statics

sigmantium:
Implementation of the following two functions

- ListStaticsNearLocationOfType( x, y, z, range, objtype, flags := 0, realm := _DEFAULT_REALM )
- ListStaticsNearLocationWithFlag( x, y, z, range, flags, realm := _DEFAULT_REALM )


Admin-Yukiko:
We like this. In fact my son and I asked for it. Odd that we would also like it huh?

PR #81, opened 2019-02-19
Hide warning gump and item drop.

nogu3ira:
Added HideWarningGump and HideWarningItem into pol.cfg. Default value is 0.
Many player macros generate these messages on the console, causing flooding, especially if you have lots of players online. You can enable setting `HideWarningGump=1` and `HideWarningItem=1` in the pol.cfg


frozenblit:
Thanks :D

PR #84, opened 2019-03-09
Kiko dev

Admin-Yukiko:
Changed: pol.cfg settings HideWarningGump and HideWarningItem to ShowWarningGump and ShowWarningItem. This positive logic is consistent with all other console output settings.
The default is zero (0). The output will not be displayed unless set to one (1).


Admin-Yukiko:
Forgot to run corechanges.py before the previous commit.

PR #94, opened 2019-07-17
Change coord.z to unsigned

krequena:
Having this as signed prevented by compiler (g++ 8.3.0) to build the code.

This change fixes the compiling process in my lab.


krequena:
After seeing the CI compilation failing, seems gcc manages errors in a different way.

PR #128, opened 2020-04-18
Added ShowWarningCursor to Pol.cfg

Tercioo:
When true, to won't show the warning 'used out of sequence cursor'.


turleypol:
Should be fixed with #129

PR #130, opened 2020-04-23
Vagrant devmachine: Use gcc 5 and cmake 3.17.1

eric-swanson:
Make it so `build_linux.sh` works on the Vagrant devmachine.

I tried cmake 3.4.1, but it exited with an error about an `if` and a `GREATER_THAN`

Verification steps, assuming a clean workspace (nothing leftover from an earlier cmake run):
```
polserver $ vagrant up
polserver $ vagrant ssh

vagrant@vagrant-ubuntu-trusty-64:~$ cd polserver/bin-build
vagrant@vagrant-ubuntu-trusty-64:~/polserver/bin-build$ ./build_linux.sh
…
[100%] Built target pol
vagrant@vagrant-ubuntu-trusty-64:~/polserver/bin-build$
```

### DIFF
--- a/devmachine/prepareMachine.sh
+++ b/devmachine/prepareMachine.sh
@@ -68,13 +68,6 @@ sudo wget -O - http://llvm.org/apt/llvm-snapshot.gpg.key | sudo apt-key add -
 
 echo
 echo -------------------------------------------------------------------------------
-echo add cmake repro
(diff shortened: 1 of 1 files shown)


frozenblit:
Apparently new if operators were introduced in cmake 3.7. We should update the docs.

And it's funny to see "first-time contributor" next to your name ;)

PR #139, opened 2020-05-02
Add minimal support for ANTLR, and grammar

eric-swanson:
## Current State of This PR
- builds and runs locally with VS2019 and with vagrant build (in addition to CI)

## Overview
The intent of this PR is to add the ANTLR runtime and just enough code that it links and runs.

## Changes
1. Adds the `runtime/Cpp/runtime/src` portion of https://github.com/antlr/antlr4/archive/4.8.tar.gz to lib/antlr
- this is 1.7 MB of source code
- this seems better than including the .tar.gz file (4.2 mb binary) or requiring it to be downloaded and extracted
- it also means IDEs can generally easily navigate to the source code 

2. Adds:
- an `ecompile` command-line option `-g`
- an `ecompile.cfg` option: `DualCompileWithAntlrGrammar` (default false)

Effect: If true, passes the given .src file through the `EscriptLexer` and `EscriptParser`.

This is enough to link and call the ANTLR runtime.

I didn't document these in core-changes.txt because they are not meant to be front-facing.

3. Adds `lib/EscriptGrammar`, with the 0.1.2 grammar from the https://github.com/polserver/escript-antlr4 repo and code generated from it, in the `EscriptGrammar` namespace.

4. Includes the `EscriptGrammar/*.(cpp|h)` files directly in bscript/CMakeSources.cmake


turleypol:
The joy of static libs ;)
I think we need to modify the cmake build commands similar to the eg curl ones:
Add to the ExternalProject_Add in the CMAKE_CACHE_ARGS 
-DCMAKE_USER_MAKE_RULES_OVERRIDE=${CMAKE_CURRENT_LIST_DIR}/c_flag_overrides.cmake

turleypol:
Linux fails in the configure step due to a case missmatch. In the escript_grammar.cmake the folder is named "EScriptGrammar" but the folder is named "EscriptGrammar"

Issue #163, opened 2020-07-23
Can start multiple text entry scripts that conflict with the singleton game data information


KevinEady:
https://github.com/polserver/polserver/blob/12d6186d10422c3ad4313ca9ac0ca3cb3da47105/pol-core/pol/module/uomod2.cpp#L1572 we don't check if chr->client->gd->textentry_uoemod already exists ... like we do in eg. SendInstaResDialog() https://github.com/polserver/polserver/blob/12d6186d10422c3ad4313ca9ac0ca3cb3da47105/pol-core/pol/module/uomod2.cpp#L2133

I'm able to start multiple scripts that each call SendTextEntryGump()s and only the last one gets revived on dialog response, even though there are two dialog gumps open on the client
https://gist.github.com/KevinEady/aca6f52be001b0620b0821607a74bccb this starts multiple scripts with text entries

PR #164, opened 2020-07-23
Ensure textentry_uoemod is not overwritten

KevinEady:
Fixes: #163


pol-core/pol/module/uomod2.cpp
```diff
@@ -1589,6 +1589,9 @@ BObjectImp* UOExecutorModule::mf_SendTextEntryGump()
     return new BError( "No client attached" );
   }
 
+  if ( chr->client->gd->textentry_uoemod != nullptr )
```
frozenblit, line 1592: There are a few methods in Character:

has_active_prompt()
has_active_gump()
etc...

Maybe this would fit better as a similar method?

PR #173, opened 2020-08-01
Checkskill attrname

turleypol:
Checkskill accepts now not only the skillid but also as string the attribute name.
It's solved in general but the method is currently only used by CheckSkill.

### DIFF
--- a/pol-core/pol/uoexec.cpp
+++ b/pol-core/pol/uoexec.cpp
@@ -750,16 +750,45 @@ bool UOExecutor::getObjtypeParam( unsigned param, const Items::ItemDesc*& itemde
 
 bool UOExecutor::getSkillIdParam( unsigned param, USKILLID& skillid )
 {
-  int skillval;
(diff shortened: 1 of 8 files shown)


pol-core/pol/uoexec.cpp
```diff
@@ -750,16 +750,58 @@ bool UOExecutor::getObjtypeParam( unsigned param, const Items::ItemDesc*& itemde
...
+      return false;
+    for ( unsigned short i = 0; i <= networkManager.uoclient_general.maxskills; ++i )
+    {
+      const UOSkill& uoskill = GetUOSkill( i );
+      if ( uoskill.pAttr == attr )
+      {
+        skillid = static_cast<USKILLID>( uoskill.skillid );
+        return true;
+      }
+    }
```
eric-swanson, line 791: Could we do this lookup once when loading the attribute data (set on a field on Attribute), rather than on every getSkillIdParam (by attrname) call?

turleypol: Yes true, it shouldn't really matter since not many attributes exists, but it also doesn't hurt to store the extra information.

PR #176, opened 2020-08-02
Add exit_code to uo::Shutdown()

KevinEady:


pol-core/clib/esignal.cpp
```diff
@@ -15,7 +15,24 @@ namespace Pol
 {
 namespace Clib
 {
-std::atomic<bool> exit_signalled( false );
+std::atomic<bool> signal( false );
+std::atomic<int> code = 0;
+
+void signal_exit( int exit_code )
+{
+  code = exit_code, signal = true;
```
eric-swanson, line 23: Is saving 1 line of code worth having to think about what is going on with the comma operator here? (and in what order things happen)
```suggestion
  signalled_exit_code = exit_code;
  exit_signalled = true;
```

pol-core/clib/esignal.cpp
```diff
@@ -15,7 +15,24 @@ namespace Pol
 {
 namespace Clib
 {
-std::atomic<bool> exit_signalled( false );
+std::atomic<bool> signal( false );
+std::atomic<int> code = 0;
```
eric-swanson, line 19: I'd recommend _not_ wrapping the accessors in methods.  `exit_signalled` is checked a lot in the client I/O loops, for example, and semantically it's just a boolean that something might set.

Maybe those who are more up-to-date on C++ and atomics and will chime in. 

```suggestion
std::atomic<bool> exit_signalled( false );
std::atomic<int> signalled_exit_code( 0 );
```

pol-core/clib/esignal.cpp
```diff
@@ -15,7 +15,24 @@ namespace Pol
...
+
+bool exit_signalled()
+{
+  return signal;
+}
+
+int exit_code()
+{
+  return code;
+}
```
eric-swanson, line 34: Suggest reading these directly rather than having trivial accessors (that then need to have slightly different names)
```suggestion
```

pol-core/support/scripts/uo.em
```diff
@@ -395,7 +395,7 @@ SetRegionLightLevel( regionname, lightlevel );
 SetRegionWeatherLevel( region, type, severity, aux := 0, lightoverride := -1);
 SetScriptController( who );
 SingleClick( who, what );
-Shutdown();
+Shutdown( exitCode := 0 );
```
eric-swanson, line 398: Strongly recommend increasing the .ecl version.  If any script that calls Shutdown isn't recompiled, the script will be killed rather than calling the Shutdown method.

KevinEady: I thought https://github.com/polserver/polserver/pull/176/files#diff-be892f55676b7e6d68bfc47ff7710570R4787 would cover optional arguments?

pol-core/clib/esignal.cpp
KevinEady: funny how a typo can actually compile and still work fine... 🙃

pol-core/support/scripts/uo.em
eric-swanson: It turns out it won't, because it won't get that far.

Here's what would happen:
1. Previously-compiled script passes 0 arguments to Shutdown
2. `Executor::ins_func` [calls getParams](https://github.com/polserver/polserver/blob/master/pol-core/bscript/executor.cpp#L2272) with the number of parameters for the module function (1)
3. `Executor::getParams` [sees an empty value stack](https://github.com/polserver/polserver/blob/master/pol-core/bscript/executor.cpp#L181) and kills the script.

pol-core/clib/esignal.cpp
KevinEady: Addressed.

pol-core/clib/esignal.cpp
KevinEady: Addressed by removing the getter wrappers.

pol-core/clib/esignal.cpp
KevinEady: Addressed by removing the getter wrappers.

pol-core/support/scripts/uo.em
KevinEady: Addressed by increasing `ESCRIPT_FILE_VER_CURRENT` to `0x10`

pol-core/pol/module/uomod.cpp
```diff
@@ -4783,7 +4782,14 @@ bool UOExecutorModule::is_reserved_to_me( Item* item )
 
 BObjectImp* UOExecutorModule::mf_Shutdown()
 {
-  Clib::exit_signalled = true;
+  int exit_code;
+
+  if ( !getParam( 0, exit_code ) )
+  {
+    exit_code = 0;
+  }
```
eric-swanson, line 4790: this is the only piece I'm iffy about: what are the scenarios where `getParam()` return false, and what would we want this function to do.
- no parameter at all: shouldn't happen (compiler should always pass an argument)
- passed a different type, say a string: will exit 0 (success).  Is that what we want?

eric-swanson, line 4790: I was wrong.  The executor uses the argument count from the module function declaration in the .ecl file, so we don't need to bump the ecl version.

```suggestion
  int exit_code = 0;

  getParam( 0, exit_code );

```

pol-core/bscript/filefmt.h
```diff
@@ -35,14 +35,15 @@ namespace Bscript
 #define ESCRIPT_FILE_VER_000C 0x000C
 #define ESCRIPT_FILE_VER_000D 0x000D
 #define ESCRIPT_FILE_VER_000F 0x000F /*unicode*/
+#define ESCRIPT_FILE_VER_0010 0x0010 /* Add exit_code to os::Shutdown() */
```
eric-swanson, line 38:
```suggestion
```

pol-core/bscript/filefmt.h
```diff
@@ -35,14 +35,15 @@ namespace Bscript
...
 
 /*
     NOTE: Update ESCRIPT_FILE_VER_CURRENT when you make a
     new escript file version, to force recompile of scripts
     and report this to users when an older compiled version
     is attempted to be executed - TJ
     */
-#define ESCRIPT_FILE_VER_CURRENT ( ESCRIPT_FILE_VER_000F )
+#define ESCRIPT_FILE_VER_CURRENT ( ESCRIPT_FILE_VER_0010 )
```
eric-swanson, line 46:
```suggestion
#define ESCRIPT_FILE_VER_CURRENT ( ESCRIPT_FILE_VER_000F )
```

pol-core/pol/module/uomod.cpp
```diff
@@ -4783,7 +4782,14 @@ bool UOExecutorModule::is_reserved_to_me( Item* item )
 
 BObjectImp* UOExecutorModule::mf_Shutdown()
 {
-  Clib::exit_signalled = true;
+  int exit_code;
+
+  if ( !exec.hasParams(1) || !getParam( 0, exit_code ) )
+  {
+    exit_code = 0;
+  }
```
eric-swanson, line 4790: I feel like I have to think a lot about whether `exit_code` will always be initialized.  So a suggestion:

```suggestion
  int exit_code = 0;

  if ( exec.hasParams(1) )
  {
    getParam( 0, exit_code );
  }
```

Also, we are effectively ignoring the error code of `getParam`, and this calls this out.

PR #177, opened 2020-08-03
Move Compiler.verbosity_level_ to CompilerConfig.VerbosityLevel

eric-swanson:
Now settable through ecompile.cfg `VerbosityLevel`

The new compiler needs access to the variable soon-to-be-formerly-known-as `Compiler::verbosity_level_`, so this PR moves it to CompilerConfig.

### DIFF
--- a/pol-core/ecompile/ECompileMain.cpp
+++ b/pol-core/ecompile/ECompileMain.cpp
@@ -100,7 +100,6 @@ int debug = 0;
 bool quiet = false;
 bool opt_generate_wordlist = false;
 bool keep_building = false;
-bool verbose = false;
 bool force_update = false;
 bool show_timing_details = false;
 bool timing_quiet_override = false;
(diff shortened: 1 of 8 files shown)


pol-core/ecompile/ECompileMain.cpp
```diff
@@ -466,7 +466,7 @@ int readargs( int argc, char** argv )
         vlev = atoi( &argv[i][2] );
         if ( !vlev )
           vlev = 1;
-        Legacy::Compiler::setVerbosityLevel( vlev );
+        compilercfg.VerbosityLevel = vlev;
```
frozenblit, line 469: The "-v" flag also enables the variable "verbose" which is used in compile_file(). Did you leave that one out on purpose?

eric-swanson: No, I missed that.  I will take another look.  Thanks!

eric-swanson: I removed the `verbose` flag and instead check `compilercfg.VerbosityLevel`.

PR #182, opened 2020-08-04
Two new members for polcore()

AsYlum-:
- polcore().last_character_serial returns last character serial assigned by core
- polcore().last_item_serial returns last item serial assigned by core

Those two properties are already saved to data/pol.txt but where was no way to read them from escript.

Tell me what do you think. Worth merging?


frozenblit:
Looks good, thanks!

PR #191, opened 2020-08-06
poltool is now able to generate clientdata

turleypol:
* added simple test for uoconvert
* fixed poltool mapdump argument handling

Start of a series with the goal to have a CI integration of Pol.
The current generated files are more or less a demo.

### DIFF
--- a/pol-core/poltool/PolToolMain.cpp
+++ b/pol-core/poltool/PolToolMain.cpp
@@ -3,7 +3,6 @@
 #include <fstream>
 #include <string>
 
-#include <format/format.h>
 #include "../clib/Program/ProgramMain.h"
 #include "../clib/clib_endian.h"
 #include "../clib/fileutil.h"
@@ -16,6 +15,8 @@
 #include "../plib/maptile.h"
 #include "../plib/maptileserver.h"
 #include "../plib/realmdescriptor.h"
+#include "testfiles.h"
+#include <format/format.h>
 #ifdef USE_SYSTEM_ZLIB
 #include <zlib.h>
 #else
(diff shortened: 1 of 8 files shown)


pol-core/poltool/PolToolMain.cpp
```diff
@@ -42,34 +43,41 @@ void PolToolMain::showHelp()
...
 {
   short wxl = 5485, wxh = 5500, wyl = 0, wyh = 30;
   const std::vector<std::string>& binArgs = programArgs();
-
   std::string realmname = "britannia";
-  if ( binArgs.size() >= 6 )
+  if ( binArgs.size() >= 7 )
   {
-    realmname = binArgs[5];
+    realmname = binArgs[6];
```
eric-swanson, line 64: Since this changes the parameter list of this command, should we have a note in core-changes?

pol-core/poltool/PolToolMain.cpp
```diff
@@ -42,34 +43,41 @@ void PolToolMain::showHelp()
               << "            x1 y1 [x2 y2 realm]       writes polmap info to polmap.html\n"
```
eric-swanson, line 43: was this already out of date?

pol-core/poltool/PolToolMain.cpp
turleypol: It is broken since mmmh ~3years. Noone ever complained about it. I would silently fix it ;)

pol-core/poltool/PolToolMain.cpp
turleypol: Do you mean due to the argument index change?  binargs contains the complete argc thus also the binary name which wasn't considered correctly.

eric-swanson: Ah yes because mapdump is [1]

pol-core/poltool/PolToolMain.cpp
frozenblit: I'm sure someone will ask "Oh, what is this poltool? Is that new?" ;)

pol-core/poltool/PolToolMain.cpp
turleypol: If I remember it long enough, we should fix it. That the vector programargs contains also the binary name isn't something I would expect. ;)

PR #193, opened 2020-08-07
Add compiler/ Profile and Report

eric-swanson:
Add `compiler/Profile`:
- will hold a bunch of `atomic<>` for elapsed times, cache hits/misses, and so forth.
- maybe should have called `Metrics`, or something else?

Add `compiler/Report`:
- tracks warning and error count for a given script compilation, and provides methods for displaying them with context.

### DIFF
--- a/pol-core/bscript/compiler/Compiler.cpp
+++ b/pol-core/bscript/compiler/Compiler.cpp
@@ -1,5 +1,11 @@
 #include "Compiler.h"
 
+#include "../clib/fileutil.h"
+#include "../clib/logfacility.h"
+
+#include "Report.h"
+#include "compilercfg.h"
+
 namespace Pol
 {
 namespace Bscript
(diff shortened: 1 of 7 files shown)


pol-core/bscript/compiler/Compiler.cpp
```diff
@@ -32,10 +41,42 @@ void Compiler::write_included_filenames( const std::string& /*pathname*/ )
 {
 }
 
-bool Compiler::compile_file( const std::string& /*filename*/,
-                             const LegacyFunctionOrder* /*legacy_function_order*/ )
+bool Compiler::compile_file( const std::string& filename,
+                             const LegacyFunctionOrder* legacy_function_order )
 {
-  return false;
+  bool success;
```
frozenblit, line 47: I'd initialize the variable already as false instead of setting it in the catch clause. But I'm aware it's a matter of taste.
```suggestion
  bool success = false;
```

pol-core/bscript/compiler/Compiler.cpp
```diff
@@ -32,10 +41,42 @@ void Compiler::write_included_filenames( const std::string& /*pathname*/ )
...
+
+    compile_file_steps( pathname, legacy_function_order, report );
+    display_outcome( pathname, report );
+
+    bool have_warning_as_error = report.warning_count() && compilercfg.ErrorOnWarning;
+    success = !report.error_count() && !have_warning_as_error;
+  }
+  catch ( std::exception& ex )
+  {
+    ERROR_PRINT << ex.what();
```
frozenblit, line 62: Missing a newline?
```suggestion
    ERROR_PRINT << ex.what() << '\n';
```

pol-core/bscript/compiler/Report.cpp
```diff
@@ -0,0 +1,42 @@
...
+  ++errors;
+  ERROR_PRINT << source_location << ": error: " << msg;
+}
+
+void Report::report_warning( const SourceLocation& source_location, const char* msg )
+{
+  ++warnings;
+  ERROR_PRINT << source_location << ": warning: " << msg;
+}
+
```
frozenblit, line 29: Is "msg" always ending in a newline or you want to enforce that here?

frozenblit: And for a second I thought `display_warnings` was being ignored, but saw the `error()` and `warning()` methods in the include. ;)

turleypol: Why not use std::string as type of msg, const char* always leads to ugly c_str() calls

eric-swanson: Yeah, `msg` is intended to always end in a newline.  This follows the pattern. of `compiler_error` and `compiler_warning`, most call site of which end in a newline.

pol-core/bscript/compiler/Compiler.cpp
eric-swanson: Fine with me!

pol-core/bscript/compiler/Compiler.cpp
eric-swanson: Thanks!

pol-core/bscript/compiler/Report.cpp
eric-swanson: The `const char*` passed to these private methods always comes from a `fmt::Writer`, for example Report::error or Report::warning.  What's the preferred way of writing the buffer from a `fmt::Writer` to ERROR_PRINT or INFO_PRINT?

PR #210, opened 2020-08-11
Scripttest inside pol

turleypol:
* poltool is now able to generate the bare distro (baredistro argument)
* * python script to convert baredistro repository to cpp
*poltool can generate missing files to be able to start (testenv)
* * will also generate client testfiles and bare distro
* At the moment only very few things are tested, I need help ;)

CI integration will also upload the pol.log file for debugging purposes.

`ctest --output-on-failure -E escript`
Runs this test only


testsuite/pol/scripts/chartests.inc
```diff
@@ -0,0 +1,61 @@
...
+    success:=0;
+    syslog("Failed to equip backpack "+backpack+" "+res);
+    return;
+  endif
+  
+  backpack.damage_cold_mod:=10;
+  if (char.damage_cold != 10)
+    syslog("failed to apply equipped aos prop");
+    success:=0;
+    return;
```
eric-swanson, line 50: i suppose this could be `return 0;` since `return` by itself is actually (IIRC) the equivalent of `return "";`

turleypol: Fixed :)

PR #211, opened 2020-08-11
Add instruction generator, emitter, emit progend

eric-swanson:
Adds [InstructionGenerator](https://github.com/polserver/polserver/blob/ecompile-antlr/pol-core/bscript/compiler/codegen/InstructionGenerator.cpp#L719): knows which instructions to generate for high-level language constructs

Adds [InstructionEmitter](https://github.com/polserver/polserver/blob/ecompile-antlr/pol-core/bscript/compiler/codegen/InstructionEmitter.cpp#L244): knows how to emit the instructions

Also in this PR: Initialize the data and emit the `PROGEND` instruction.  This means that the compiler can now compile an empty file to valid .ecl.

```
$ touch a.src
$ /home/vagrant/polserver/bin/ecompile -l -g a.src
$ cat a.lst
0: progend
```

### DIFF
--- a/pol-core/bscript/compiler/codegen/CodeGenerator.h
+++ b/pol-core/bscript/compiler/codegen/CodeGenerator.h
@@ -22,6 +22,7 @@ private:
   void generate_instructions( CompilerWorkspace& );
 
 private:
+  // Verb vs noun - see InstructionGenerator for reasoning
   InstructionEmitter& emitter;
   InstructionEmitter& emit;
 };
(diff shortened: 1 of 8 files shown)


pol-core/bscript/compiler/codegen/CodeGenerator.h
```diff
@@ -18,9 +17,13 @@ class CodeGenerator
...
 
 private:
-  CodeGenerator();
+  explicit CodeGenerator( InstructionEmitter& );
 
   void generate_instructions( CompilerWorkspace& );
+
+private:
+  InstructionEmitter& emitter;
+  InstructionEmitter& emit;
```
eric-swanson, line 26: There are two of these because sometimes when calling a method on `InstructionEmitter`, the variable name reads better as a noun, and sometimes it reads better as a verb.

[Example.](https://github.com/polserver/polserver/blob/ecompile-antlr/pol-core/bscript/compiler/codegen/InstructionGenerator.cpp#L667)

PR #212, opened 2020-08-12
Update grammar

eric-swanson:
Update Lexer and Parser g4 definitions:
   - Disallow expression without semicolon at end of file
   - Add "unambiguous" grammar (alternative to original left-recursive grammar)
     - this was needed to allow proper operator precedence with `.+` `.?` `.-` in combination with `:=`
   - Allow `=` to parse, so that we can report an error
   - Add Elvis operator
   - Allow both "byref" and "unused" in same function parameter

It will probably be easier to review the [first commit](https://github.com/polserver/polserver/pull/212/commits/763d9ee7f77a2669b8d1fd8df18f4004e3a64dea) separately (the changes to the .g4 files) from the second (the regenerated sources)


eric-swanson:
I'm moving this PR to a later commit on the branch.  Maybe we can avoid the "unambiguous grammar" altogether.  Right now there are still some scripts in the ClassicDistro that the left-recursive grammar does not compile.

Leaving this as an open draft for discussion in the meantime.

PR #215, opened 2020-08-13
item.hitscript and item.onhitscript will now return the full spec for the script. Added breaking-changes.txt.

frozenblit:
breaking-changes.txt should be kept short. And we can add more as we find them.

### DIFF
--- a/pol-core/pol/scrdef.cpp
+++ b/pol-core/pol/scrdef.cpp
@@ -101,7 +101,10 @@ bool ScriptDef::config_nodie( const std::string& iname, const Plib::Package* ipk
 
 std::string ScriptDef::qualifiedname() const
 {
-  return relativename( nullptr );
+  if ( empty() )
+    return "";
+  else
+    return ":" + ( pkg_ ? pkg_->name() : "" ) + ":" + localname_.get();
 }
 
 std::string ScriptDef::relativename( const Plib::Package* pkg ) const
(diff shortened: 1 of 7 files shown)


pol-core/pol/uoscrobj.cpp
```diff
@@ -4396,7 +4396,8 @@ BObjectImp* UWeapon::get_script_member_id( const int id ) const
     return new String( attribute().name );
     break;
   case MBR_HITSCRIPT:
-    return new String( hit_script_.relativename( tmpl->pkg ) );
+    // bad method name? nullptr makes it return the fullpath
+    return new String( hit_script_.relativename( nullptr ) );
```
eric-swanson, line 4400: How about we make a new method called `fullname` or `absolutename` or `package_qualified_name` ?

eric-swanson: (it can call relativename(nullptr) )

pol-core/pol/scrdef.cpp
```diff
@@ -99,9 +99,14 @@ bool ScriptDef::config_nodie( const std::string& iname, const Plib::Package* ipk
   return true;
 }
 
+std::string ScriptDef::qualifiedname() const
+{
+  return relativename( nullptr );
```
eric-swanson, line 104: I think if pkg_ is null, this will return the unqualified name, rather than `::x/y/z`.

Suggestion to fix:
```suggestion
  if ( name_ == "" )
    return "";
  else
    return ":" + ( pkg_ ? pkg_->name() : "" ) + ":" + localname_.get();
```

PR #219, opened 2020-08-14
Cmake and zlib cleanup

frozenblit:
Mostly 3 changes:
- The path to "lib/" is now set explicitly instead of relying on relative paths.
- use_zlib() takes care of linking to zlib on all systems and includes zlib to the include path on Windows
- The relative paths to zlib are gone


pol-core/pol/module/uomod2.cpp
```diff
@@ -102,11 +102,8 @@
 #endif
 
 
-#ifdef USE_SYSTEM_ZLIB
```
eric-swanson, line 105: nice

PR #222, opened 2020-08-15
Add file/SourceFileCache

eric-swanson:
There is one for *.em files (which is meant to be large enough that they are loaded once and never evicted), and another for *.inc files.

### DIFF
--- a/pol-core/bscript/compiler/file/SourceFileCache.cpp
+++ b/pol-core/bscript/compiler/file/SourceFileCache.cpp
@@ -1,0 +1,95 @@
+#include "SourceFileCache.h"
+
+#include "clib/logfacility.h"
+#include "compiler/Profile.h"
+#include "compiler/file/SourceFile.h"
+#include "compiler/file/SourceFileIdentifier.h"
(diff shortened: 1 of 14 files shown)


pol-core/bscript/compiler/file/SourceFileCache.cpp
```diff
@@ -0,0 +1,95 @@
...
+    return;
+
+  std::unique_lock<std::mutex> lock( mutex, std::try_to_lock_t() );
+  if ( !lock.owns_lock() )
+    return;
+
+  if ( files.size() <= keep )
+    return;
+
+  std::vector<SourceFileAndFrequency> pathname_frequencies;
```
frozenblit, line 73: It would be good to reserve the capacity of the vector here
```suggestion
  std::vector<SourceFileAndFrequency> pathname_frequencies;
  pathname_frequencies.reserve(files.size());
```

eric-swanson: Thanks!

PR #235, opened 2020-08-20
New compiler: Add StringValue

eric-swanson:
The meat of ValueBuilder::unquote is adapted (but mostly copied) from [Parser::tryLiteral](https://github.com/polserver/polserver/blob/master/pol-core/bscript/parser.cpp#L1251).

```
$ cat a.src
"xyz";
$ /home/vagrant/polserver/bin/ecompile -l -g a.src
$ cat a.lst
0: "xyz" (string) len=3 offset=0x1
1: # (consume)
2: progend
```

### DIFF
--- a/pol-core/bscript/compiler/astbuilder/ValueBuilder.cpp
+++ b/pol-core/bscript/compiler/astbuilder/ValueBuilder.cpp
@@ -1,5 +1,7 @@
 #include "ValueBuilder.h"
 
+#include <cstring>
+
 #include "clib/strutil.h"
 #include "compiler/Report.h"
 #include "compiler/ast/FloatValue.h"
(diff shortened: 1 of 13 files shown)


pol-core/bscript/compiler/astbuilder/ValueBuilder.cpp
```diff
@@ -23,10 +27,97 @@ std::unique_ptr<FloatValue> ValueBuilder::float_value(
...
+
+std::string ValueBuilder::unquote( antlr4::tree::TerminalNode* string_literal )
+{
+  std::string input = string_literal->getSymbol()->getText();
+  const char* s = input.c_str();
+  if ( *s != '\"' )
+    location_for( *string_literal ).internal_error( "string does not begin with a quote?" );
+
+  const char* end = s + 1;
+  std::string lit;
```
frozenblit, line 45: This function could use some refactoring later and I'm aware it's the same code as the old compiler. Should we at least reserve the string size?
```suggestion
  std::string lit;
  lit.reserve(input.length());
```

frozenblit (approved):

PR #236, opened 2020-08-20
Upgrade Vagrant setup to Ubuntu 20.04 (focal64)

eric-swanson:
Another note: with this setup, if you increase the cpu count and memory, it builds much faster.  Example: with cpus = 10, memory = 16384, for me it buildtin ~2.5 minutes rather than 18-20.

Verified on a new clone:
```
$ vagrant up
$ vagrant ssh
$ cd polserver/bin-build
$ ./build_linux.sh
```

On an existing clone, you'll need to clean, as well as track down and wipe out any already-built external libraries.

### DIFF
--- a/devmachine/prepareMachine.sh
+++ b/devmachine/prepareMachine.sh
@@ -5,7 +5,17 @@
 
 # Updates the system and install needed packages
 # uses non-interactive mode (http://askubuntu.com/questions/146921/how-do-i-apt-get-y-dist-upgrade-without-a-grub-config-prompt)
-sudo apt-get update
+
+echo
+echo -------------------------------------------------------------------------------
+echo apt-get update
+echo -------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


devmachine/prepareMachine.sh
```diff
@@ -46,69 +56,37 @@ sudo apt-get install libssl-dev -y
...
+echo debugging
 echo -------------------------------------------------------------------------------
-echo
-sudo apt-get install pkg-config uuid-dev
-
-
-sudo apt-get upgrade
+sudo apt-get install gdb valgrind -y
 
 # Creates a link to /vagrant (which, if you run this file, will contain the SVN repository)
```
frozenblit, line 91: ;)
```suggestion
# Creates a link to /vagrant (which, if you run this file, will contain the git repository)
```

frozenblit (approved):

frozenblit (approved):

PR #243, opened 2020-08-26
New compiler: Add IntegerValue

eric-swanson:
The last two PRs were pretty large, so here is a smaller one.

This adds support for integer values.

The grammar actually defines negative integer values as `unary operator -` followed by `positive integer literal`.  The next PRs will add the UnaryOperator and enough of the optimizer to convert such sequences to an IntegerValue with a negative value.

### DIFF
--- a/pol-core/bscript/compiler/astbuilder/ValueBuilder.cpp
+++ b/pol-core/bscript/compiler/astbuilder/ValueBuilder.cpp
@@ -148,7 +148,7 @@ int ValueBuilder::to_int( EscriptParser::IntegerLiteralContext* ctx )
   }
   else if ( auto hex_literal = ctx->HEX_LITERAL() )
   {
-    return (int)std::stoul( hex_literal->getSymbol()->getText(), nullptr, 16 );
+    return static_cast<int>( std::stoul( hex_literal->getSymbol()->getText(), nullptr, 16 ) );
   }
   else if ( auto oct_literal = ctx->OCT_LITERAL() )
   {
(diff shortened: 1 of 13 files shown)


pol-core/bscript/compiler/astbuilder/ValueBuilder.cpp
```diff
@@ -128,4 +140,25 @@ std::unique_ptr<Value> ValueBuilder::value( EscriptParser::LiteralContext* ctx )
...
 
+int ValueBuilder::to_int( EscriptParser::IntegerLiteralContext* ctx )
+{
+  if ( auto decimal_literal = ctx->DECIMAL_LITERAL() )
+  {
+    return std::stoi( decimal_literal->getSymbol()->getText() );
+  }
+  else if ( auto hex_literal = ctx->HEX_LITERAL() )
+  {
+    return (int)std::stoul( hex_literal->getSymbol()->getText(), nullptr, 16 );
```
frozenblit, line 151: I'd rather make it ugly here so the cast sticks out:
```suggestion
    return static_cast<int>(std::stoul( hex_literal->getSymbol()->getText(), nullptr, 16 ));
```
Is the old compiler already handling hex as unsigned?

frozenblit (approved):

pol-core/bscript/compiler/astbuilder/ValueBuilder.cpp
eric-swanson: Thanks, my oversight

And yeah, it does compile this
```
$ cat a.src
use math;
print(hex(0x90000000));

$ /home/vagrant/polserver/bin/ecompile -l a.src
EScript Compiler v1.15
Copyright (C) 1993-2018 Eric N. Swanson

Compiling: a.src
Writing:   a.ecl
Writing:   a.lst

$ /home/vagrant/polserver/bin/runecl a.ecl
EScript Executor v1.15
Copyright (C) 1993-2018 Eric N. Swanson

0x90000000
```

PR #272, opened 2020-09-05
New compiler: Add support for case statements

eric-swanson:
Adds:
- ast/CaseDispatchDefaultSelector: AST node for a "default:" selector
- ast/CaseDispatchGroup: AST node for a collection of selectors and the code to execute
- ast/CaseDispatchGroups: AST node holding all groups in a case statement
- ast/CaseDispatchSelectors: AST node holding all selectors for one group
- ast/CaseStatement: AST node for a whole case statement
- codegen/CaseDispatchGroupVisitor: knows what to put in a case jump dispatch table
- codegen/CaseJumpDataBlock: knows how to build the entries in the data block

```
case (expr)
    // dispatch group 0:
    4:              // an integer selector
    "xy":           // a string selector
      some_code();

    // dispatch group 1:
    A_CONST:
    default:        // a default selector
      default_stuff();
endcase
```


pol-core/bscript/compiler/astbuilder/CompoundStatementBuilder.cpp
```diff
@@ -87,6 +98,62 @@ std::unique_ptr<Block> CompoundStatementBuilder::block( EscriptParser::BlockCont
...
+        selectors.push_back(
+            std::make_unique<Identifier>( location_for( *identifier ), text( identifier ) ) );
+      }
+      else if ( auto string_literal = group_label->STRING_LITERAL() )
+      {
+        selectors.push_back( string_value( string_literal ) );
+      }
+      else if ( auto dflt = group_label->DEFAULT() )
+      {
+        selectors.push_back(
```
frozenblit, line 140: Do we want to at least warn if more than a default is defined?

frozenblit (approved):

pol-core/bscript/compiler/astbuilder/CompoundStatementBuilder.cpp
eric-swanson: Ah thanks, good catch.  Yes, but in the semantic analyzer, and for all of the selector types.

frozenblit: Cool! I'm now curious how many duplicate ints/strings are out there.

frozenblit (approved):

Issue #311, opened 2020-09-14
New compiler error #3 - unusual syntax produces comparison output mismatch


AsYlum-:
script:
```
program test_ecompile()
    var test;
    // some code
    test := test := "something"; // <--- this should be test := "something"; 
endprogram
```
In comparison mode that will produce compiler output mismatch but compiles fine on both compilers. Only comparison will complain.

```
Listings match: 0
Binary (.ecl) outputs match: 0
Failing because output does not match.
  - og-compiler.ecl
  - new-compiler.ecl
Execution aborted due to: Compiler output mismatch
```

PR #314, opened 2020-09-15
New compiler: warn on double assignment to the same variable

eric-swanson:
Warn on assignment statements like this, for which the OG compiler and the new compiler generate different code:

```
    x := x := expr;
```

This fixes https://github.com/polserver/polserver/issues/311

The OG compiler doesn't optimize the leftmost assignment to an 'assign-variable-consume',
so a listfile for the above looks like this:
```
  0: decl global #0
  1: #
  2: global #0
  3: global #0
  4: :=
  5: "hi"
  6: := #
  7: progend
```

The new compiler optimizes like this:
```
  0: decl global #0
  1: #
  2: global #0
  3: "hi"
  4: :=
  5: global0 :=
  6: progend
```

For reference, this is what a single assignment looks like:
```
  0: declare global #0
  1: # (consume)
  2: "hi" (string) len=2 offset=0x1
  3: global #0 :=
  4: progend
```

### DIFF
--- a/pol-core/bscript/compiler/analyzer/SemanticAnalyzer.cpp
+++ b/pol-core/bscript/compiler/analyzer/SemanticAnalyzer.cpp
@@ -2,6 +2,7 @@
 
 #include <boost/range/adaptor/reversed.hpp>
 
+#include "clib/logfacility.h"
 #include "clib/strutil.h"
 #include "compiler/Report.h"
 #include "compiler/analyzer/Constants.h"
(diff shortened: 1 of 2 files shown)


pol-core/bscript/compiler/analyzer/SemanticAnalyzer.cpp
```diff
@@ -72,6 +75,8 @@ void SemanticAnalyzer::register_const_declarations( CompilerWorkspace& workspace
 
 void SemanticAnalyzer::analyze()
 {
+  INFO_PRINT << workspace.top_level_statements->to_string_tree() << "\n";
```
frozenblit, line 78: Forgot to remove the print?

frozenblit (approved):

PR #323, opened 2020-09-23
Hoist networking members and methods of Client into a new base class

frozenblit:
Hoisted members and methods of Client responsible for networking into a new base, ThreadedClient, at the same access levels. Thus the remaining code doesn't have to change yet.

### DIFF
--- a/pol-core/pol/network/client.h
+++ b/pol-core/pol/network/client.h
@@ -153,7 +153,8 @@ protected:
   ThreadedClient( Crypt::TCryptInfo& encryption, const Client& myClient );
 
 public:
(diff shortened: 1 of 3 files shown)


pol-core/pol/network/client.h
```diff
@@ -128,7 +128,92 @@ typedef struct
...
+
+class Client;
+
+class ThreadedClient
+{
+public:
+  bool isReallyConnected() const;
+  bool isConnected() const;
+
+  // void Disconnect();
```
eric-swanson, line 140: could delete this comment since it's protected beloe
```suggestion
```

eric-swanson (approved):

PR #340, opened 2020-10-19
Upgrade boost to 1.74

eric-swanson:
- Upgrade to boost 1.74
- Build libraries needed for asio

This uses a bintray.com URL because the boost 1.74 .bz2 file exceeds the GitHub file size limit of 100 MB.


cmake/Boost.cmake
```diff
@@ -1,34 +1,71 @@
 message("* boost")
-set (BOOST_SOURCE_DIR "${POL_EXT_LIB_DIR}/boost_1_67_0")
+set (BOOST_SOURCE_DIR "${CMAKE_CURRENT_LIST_DIR}/../lib/boost_1_74_0")
```
turleypol, line 2: Should use the POL_EXT_LIB_DIR variable instead of current_list_dir

cmake/Boost.cmake
```diff
@@ -1,34 +1,71 @@
...
+  set (BOOST_BUILD_COMMAND "./b2")
+  set (BOOST_REGEX_LIB "${BOOST_STAGE_LIB_DIR}/libboost_regex.a" )
+  set (BOOST_SYSTEM_LIB "${BOOST_STAGE_LIB_DIR}/libboost_system.a" )
+  set (BOOST_THREAD_LIB "${BOOST_STAGE_LIB_DIR}/libboost_thread.a" )
 endif()
 
-set (BOOST_LIBRARY_DIR "${BOOST_SOURCE_DIR}/lib/lib")
+if(NOT EXISTS ${BOOST_REGEX_LIB} OR NOT EXISTS ${BOOST_SYSTEM_LIB} OR NOT EXISTS ${BOOST_THREAD_LIB})
+  message("  - will build in ${BOOST_SOURCE_DIR} toolset=${BOOST_TOOLSET}")
+  ExternalProject_Add(boost
```
turleypol, line 37: The boost folder should be cached in our CI like the other bigger 3rd party libs.

pol-core/pol/CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ warning_suppression(${ex_name})
 enable_pch(${ex_name})
 use_curl(${ex_name})
 use_benchmark(${ex_name})
+use_boost(${ex_name})
```
turleypol, line 17: Clib adds a dependency to the (old) boost target, to guarantee that it got extracted. Now the boost target builds boost, so nothing from the core can get build in parallel. Clib should be changed to depend on the extract target which is enough for all other Pol libs.

cmake/Boost.cmake
eric-swanson: Sounds good.  How?

turleypol: In the yml files in .GitHub/workflow is something like:
- name: Cache antlr
        uses: actions/cache@v1
        with:
          path: lib/antlr
          key: ${{ runner.os }}-${{ matrix.config.cxx-compiler }}-${{ matrix.config.compiler-version }}-antlr-buildtypefix

(And the same for curl)
Basically just copy it and adapt the name and the path. The key is something unique which (hopefully) triggers a rebuild when something changes which needs a rebuild. In this case I guess the os and compiler version is enough since the boost version is part of the directory name.

PR #344, opened 2020-10-22
Removed absolute paths from ecompil.cfg.example

Admin-Yukiko:


pol-core/support/scripts/ecompile.cfg.example
```diff
@@ -5,25 +5,26 @@
 #
 # ModuleDirectory: Location of *.em files
 #
-ModuleDirectory=D:\DistroDev\scripts\modules
+ModuleDirectory=/scripts/modules
```
eric-swanson, line 8: Aren't these technically still "absolute"?

KevinEady (changes requested):
The paths need to start with a `.`, ie. `./scripts/modules`.

pol-core/support/scripts/ecompile.cfg.example
Admin-Yukiko: If no absolute path is specified POL handles them as relative to POL root, or at least it used to. I haven't tested with Cores later than 098.

PR #345, opened 2020-10-25
Strftime

turleypol:
Updated strformattime to allow all new c++11 formats. (Except these which vs2019 does not support) removed the restriction of 100chars


pol-core/pol/module/utilmod.cpp
```diff
@@ -164,11 +224,13 @@ Bscript::BObjectImp* UtilExecutorModule::mf_StrFormatTime()
     }
   }
 
-  char buffer[102];  // +2 for the \0 termination.
-  if ( strftime( buffer, sizeof buffer, format_string->data(), &time_struct ) > 0 )
-    return new String( buffer );
-  else
-    return new BError( "Format string too long." );
+  std::string buffer;
```
frozenblit, line 227: Should you use `std::vector<char>` here instead? Non-const `buffer.data()` is new in C++17 and it comes with a [warning](https://en.cppreference.com/w/cpp/string/basic_string/data):

> 2) Modifying the past-the-end null terminator stored at data()+size() to any value other than CharT() has undefined behavior.

I guess because `str.size()` does not update when you modify `str.data()` directly. Anyway, an `std::vector<char>` would work the same way and avoid UB.

turleypol: Good point, I changed it to be on the safe side.

frozenblit (approved):

pol-core/pol/module/utilmod.cpp
```diff
@@ -164,11 +225,13 @@ Bscript::BObjectImp* UtilExecutorModule::mf_StrFormatTime()
...
-  if ( strftime( buffer, sizeof buffer, format_string->data(), &time_struct ) > 0 )
-    return new String( buffer );
-  else
-    return new BError( "Format string too long." );
+  std::vector<char> buffer;
+  buffer.reserve( std::max( format_string->length() * 2, (size_t)50u ) );
+  while ( strftime( buffer.data(), buffer.capacity(), format_string->data(), &time_struct ) == 0 )
+  {
+    buffer.reserve( buffer.capacity() * 2 );
+  }
```
eric-swanson, line 233: Should we use `resize` rather than `reserve`?

From https://en.cppreference.com/w/cpp/container/vector/data
> If size() is 0, data() may or may not return a null pointer.

frozenblit: Good catch! I guess today is UB day

eric-swanson (approved):

PR #352, opened 2020-12-01
Clean up warnings: size_t/unsigned mixing, unused private variables

eric-swanson:


pol-core/bscript/compiler/astbuilder/ExpressionBuilder.cpp
```diff
@@ -38,7 +38,7 @@ ExpressionBuilder::ExpressionBuilder( const SourceFileIdentifier& source_file_id
 {
 }
 
-[[noreturn]] BTokenId ExpressionBuilder::unhandled_operator( const SourceLocation& source_location )
```
frozenblit, line 41: I'm conflicted on this one. Without `[[noreturn]]` and with `BTokenId`, this method's signature implies that `unhandled_operator` could return a default value. Wouldn't a `[[noreturn]] void` be clearer here?

pol-core/bscript/compiler/codegen/DataEmitter.cpp
```diff
@@ -35,10 +37,13 @@ unsigned DataEmitter::store( const std::byte* data, unsigned len )
 
 unsigned DataEmitter::append( const std::byte* data, unsigned len )
 {
-  unsigned position = data_section.size();
+  size_t position = data_section.size();
+  if ( position > std::numeric_limits<unsigned>::max() ) {
```
frozenblit, line 41: Can the data still be addressed if position+len is beyond unsigned::max()? Should it perhaps check for that instead?

pol-core/bscript/compiler/codegen/InstructionEmitter.cpp
```diff
@@ -163,7 +163,7 @@ unsigned InstructionEmitter::casejmp()
 unsigned InstructionEmitter::case_dispatch_table( const CaseJumpDataBlock& dispatch_table )
 {
   auto& bytes = dispatch_table.get_data();
-  return data_emitter.append( bytes.data(), bytes.size() );
+  return data_emitter.append( bytes.data(), static_cast<unsigned>( bytes.size() ) );
```
frozenblit, line 166: This makes me wonder if append() and store() should take a size_t for length instead. Avoids the ugly cast.

frozenblit (approved):

pol-core/bscript/compiler/astbuilder/ExpressionBuilder.cpp
eric-swanson: That's the first thing I tried.  This is used in expressions where the operator doesn't match any of the expected values.  If it doesn't have a return type of BTokenId then it doesn't match the type of the expression.

So the options are:
- leave the warning (`[[noreturn BTokenId]`)
- eliminate the warning (but this means it is a function that doesn't ever return, but is not marked `[[noreturn]]` and has a non-void return type)
- redo the expressions to use if-then-else rather than the ternary chain

pol-core/bscript/compiler/codegen/DataEmitter.cpp
eric-swanson: Sure, we can go with that

pol-core/bscript/compiler/codegen/InstructionEmitter.cpp
eric-swanson: That does sound better.  Thanks!

pol-core/bscript/compiler/astbuilder/ExpressionBuilder.cpp
eric-swanson: I redid the ternary expressions and went with `[[noreturn]] void`.  Thanks!

PR #359, opened 2020-12-26
Get environment variable/Extended Testsuite

### LINKED ISSUE #141: GetEnvironmentVariable(variable := 0)
Please introduce this method in any module you seem fit. If no variable passed, return a struct of [var] := val of all env variables.

turleypol:
Closes #141 
Uses it in the Testsuite to filter tests and to be able to test the restart case (save/load)
testpkg/restart is the only pkg which will be run in the second start of Pol.


pol-core/pol/module/osmod.h
```diff
@@ -118,6 +118,7 @@ class OSExecutorModule : public Bscript::TmplExecutorModule<OSExecutorModule, Co
 
   [[nodiscard]] Bscript::BObjectImp* mf_PerformanceMeasure();
   [[nodiscard]] Bscript::BObjectImp* mf_LoadExportedScript();
+  [[nodiscard]] Bscript::BObjectImp* mf_GetEnvironmentVariable();
```
eric-swanson, line 121: This seems like a security risk.  Can we have a config file that contains a list of environment variables names that are allowed to be queried?

turleypol: Mmmh in which way is getting an environment variable a security risk?
For example in my old shard we ran two servers where special files where needed to be able to distinguish between them. Via an environment variable this would be way easier.
The same is true for SQL database password, instead of a cfg file with the password which could be by accident checked in an environment variable on the server is much safer.

eric-swanson: The SQL database password provides the best example.  If any script in any package can read any environment variable, then if a database password is present in the environment this presents a security risk, in that any time you install any package you have to be aware that it could access that database password.

turleypol: Mmmh funny attack vector.

Seems we have fundamental different opinion how scripts are shared and what intellectual abilities the shard admins have.
Added a pol.cfg setting..

docs/docs.polserver.com/pol100/osem.xml
```diff
@@ -283,4 +283,13 @@ Example: create a POST request to a REST endpoint with custom header:
     <explain>Starts a new script, to be able to later call exported functions. Non-critical or critical depending on the current state of the calling script. Blocks until the program part of the script is finished. See ExportedScript documentation how to call exported functions.</explain>
     <related>ExportedScript</related>
 </function>
+
+<function name="GetEnvironmentVariable">
+  <prototype>GetEnvironmentVariable(name:="")</prototype>
+    <parameter name="name" value="String name of environment variable, or empty string" />
+    <return>String value of given enviroment variable, or if no name is given Dictionary{String, String}</return>
+    <explain>Returns the value of given environment variable, or if no name is given it returns a dictionary with all environment variables.</explain>
```
frozenblit, line 291: "all allowed environment variables"?

pol-core/doc/core-changes.txt
```diff
@@ -1,4 +1,8 @@
 ﻿-- POL100 --
+02-26-2020 Turley, Kevin:
+    Added: os::GetEnvironmentVariable(name:="")
+           Returns String value of given environment variable name. If name is empty, returns Dictionary of all environment variables.
```
frozenblit, line 4: "all allowed"?

pol-core/pol/module/osmod.h
frozenblit: This should be solved now with the setting to whitelist environment variables added by @KevinEady

docs/docs.polserver.com/pol100/corechanges.xml
```diff
@@ -2,9 +2,16 @@
...
 		<topic>Latest Core Changes</topic>
-		<datemodified>02-22-2021</datemodified>
+		<datemodified>02-26-2021</datemodified>
 	</header>
 	<version name="POL100">
+		<entry>
+			<date>02-26-2020</date>
+			<author>Turley, Kevin:</author>
+			<change type="Added">os::GetEnvironmentVariable(name:=&quot;&quot;)<br/>
+Returns String value of given environment variable name. If name is empty, returns Dictionary of all environment variables.<br/>
```
frozenblit, line 12:
```suggestion
Returns String value of given environment variable name. If name is empty, returns Dictionary of all allowed environment variables.<br/>
```

frozenblit (commented):
Just a little change to the docs

frozenblit (approved):

pol-core/pol/module/osmod.h
karolswdev: @eric-swanson environment variables are NOT a safe place to store any secrets AT ALL, at ANY GIVEN TIME. Therefore, in my view, the argument to begin with is invalid. On the other hand, it enables you to gain incredible flexibility without too much investment in your code.

frozenblit: Hi Karol, thanks for your comments. I believe we found a good compromise. I'm marking this as resolved now.

PR #364, opened 2021-02-01
Add the option to receive only PC or NPC area events

frozenblit:
EnableEvents(ev, range, evmask)

where evmask can be:
0 -> accepts events from both PC or NPC
1 -> accepts events from PC only
2 -> accepts events from NPC only


KevinEady (approved):
LGTM! On CI too:

```
testnpc (testpkgs/npc/)
68:   Calling test_npc.ecl..
68:     Calling npc_ai_movement..
68:     Calling npc_enteredarea..
68:     Calling npc_enteredarea_npconly..
68:     Calling npc_enteredarea_pconly..
```

PR #371, opened 2021-02-24
Remove erroneous lexing productions

KevinEady:
The Escript ANTLR grammar development started  with a language that is more "full-featured" than Escript. This PR removes those additional allowed characters in the lexer.


lib/EscriptGrammar/EscriptLexer.g4
```diff
@@ -82,18 +82,16 @@ TOK_IN:             'in';
...
-FLOAT_LITERAL:      (Digits '.' Digits? | '.' Digits) ExponentPart? [fFdD]?
-             |       Digits (ExponentPart [fFdD]? | [fFdD])
+FLOAT_LITERAL:      (Digits '.' Digits? | '.' Digits) ExponentPart?
+             |       Digits ExponentPart
              ;
 
-HEX_FLOAT_LITERAL:  '0' [xX] (HexDigits '.'? | HexDigits? '.' HexDigits) [pP] [+-]? Digits [fFdD]?;
-
-CHAR_LITERAL:       '\'' (~['\\\r\n] | EscapeSequence) '\'';
+HEX_FLOAT_LITERAL:  '0' [xX] (HexDigits '.'? | HexDigits? '.' HexDigits);
```
frozenblit, line 94: This is missing the exponent part of the hex float. Do we want to leave it out?

KevinEady: You are right. When attempting to compile `0x10.1` in cpp,  you get an error:
```
test.cc:4:29: error: hexadecimal floating literal requires an exponent
    cout << "Hel " << 0x10.1;
```

So the `p` is required.

frozenblit (approved):

PR #374, opened 2021-02-24
Update breaking-changes.txt

simona70:
I thing this should be added


frozenblit:
Thanks, @simona70! We somehow thought that multilines weren't widely used because they were never officially part of the language. But we now realize that many shards relied on them.

I hope you haven't updated your scripts yet. @KevinEady is working on making the compiler accept multilines again:
https://github.com/polserver/polserver/pull/375
So it won't be a breaking change for much longer

PR #380, opened 2021-02-28
Misc testsuite

turleypol:
restartpkg will now be executed always on the end, to prevent cleanup by other pkgs
updated sysevent.inc
fixed output in escripttests when linecount differs


KevinEady (approved):
LGTM

PR #391, opened 2021-03-25
Documentation: fixed invalid parameter Hits value range for ApplyRawDamage()

### LINKED ISSUE #125: ApplyRawDamage needs changed.
ApplyRawDamage should not be limited to the range 0..255. Some "bosses" might have HP in the tens of thousands on some servers and if there is a special "boss weapon" that can deal damage greater than 255 it fails to do any damage because ApplyRawDamage will not accept a number above 255.

AsYlum-:
Fixes #125


frozenblit (approved):

frozenblit:
Thanks!

PR #393, opened 2021-03-27
Testclient api: listobjects, open backpack,  target

turleypol:
Plus cleanup

### DIFF
--- a/testsuite/testclient/pyuo/pyuo/packets.py
+++ b/testsuite/testclient/pyuo/pyuo/packets.py
@@ -158,7 +158,7 @@ class Packet():
     ''' Add a signed char (byte) to the packet '''
     if not isinstance(val, int):
       raise TypeError("Expected int, got {}".format(type(val)))
-    if val < 0 or val > 255:
+    if val < -255 or val > 255:
       raise ValueError("Byte {} out of range".format(val))
     self.buf += struct.pack('b', val)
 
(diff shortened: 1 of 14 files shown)


testsuite/testclient/pyuo/pyuo/packets.py
```diff
@@ -158,7 +158,7 @@ def eschar(self, val):
     ''' Add a signed char (byte) to the packet '''
     if not isinstance(val, int):
       raise TypeError("Expected int, got {}".format(type(val)))
-    if val < 0 or val > 255:
+    if val < -255 or val > 255:
```
frozenblit, line 161: -128 to 127?

testsuite/testclient/pyuo/testclient.py
```diff
@@ -45,21 +48,51 @@ def addTodo(self,ev):
...
-        elif todo=="move":
-          self.client.move(arg)
+    with self.todosLock:
+      if not len(self.todos):
+        return True
+      todos = self.todos.copy()
+      self.todos.clear()
+    while len(todos):
+      res = todos.popleft()
+      todo=res["todo"]
```
frozenblit, line 58: Maybe it's ok for now like this, but we should consider using a dictionary of todo <-> lambdas/methods for processing each command

frozenblit (approved):
Looks good to me ;)

PR #403, opened 2021-06-20
Smooth Boat pkt: overflow/invis objects

turleypol:
Rewrite smooth boat pkt so it does not leak info of invisible objects. Send the correct objcount when truncating.
Testclient is now version 7
New tests:
- move a boat with a client on it while more items exists then the boat pkt can handle
- move a boat and check if the client gets the correct pos of boat and containing item


KevinEady (approved):
:+1: LGTM

PR #404, opened 2021-07-20
Fixed target cancelling bug

KevinEady:


pol-core/pol/target.cpp
```diff
@@ -157,6 +138,13 @@ void TargetCursor::cancel( Mobile::Character* chr )
 
 void TargetCursor::handle_target_cursor( Mobile::Character* chr, PKTBI_6C* msg )
 {
+  u32 target_cursor_serial = cfBEu32( msg->target_cursor_serial );
+  if ( target_cursor_serial != cursorid_ )
+  {
+    SuspiciousActs::OutOfSequenceCursor( chr->client );
```
turleypol, line 144: I think CUO would then also have the suspicicousact when by script CancelTarget is called. Since it sends as cursor_serial in this case a 0.

frozenblit (approved):

AsYlum-:
Just for reference there is a fix for this in CUO client https://github.com/andreakarasho/ClassicUO/issues/1373

PR #405, opened 2021-07-21
Boat pkt encoding len

turleypol:
Huffman encoding can need more space then the uncompressed pkt.
This leads to a crash in the newer boat pkts. 
The fix is simple: number of maximum items is now reduced so that the uncompressed pkt has a maximum size of 0xf000.


frozenblit (approved):
Looks good to me

PR #406, opened 2021-07-30
Descriptorbugs

turleypol:
Fixed: loading of npcdesc AttackCProp and ShieldCProp
Fixed: loading of itemdesc NoDrop and NoDropException
Changed: either AttackSpeed or AttackDelay npcdesc entry need to exist for npc weapon creation.
Before only AttackSpeed lead to an creation.


KevinEady (approved):
🔥

PR #412, opened 2021-08-22
Unittests

turleypol:
Merged the existing two unittests which only prints with the pol -test functionality. Integrated it into CI.


frozenblit (approved):
Looks good to me

PR #416, opened 2021-09-01
Initial pos class

turleypol:
Added unused position classes plus tests


frozenblit (approved):
Looks good. I'd still make the include paths relative to the project root. You might also want to rename InRange / CropX / CropY to be consistent with the other method names (in_range, crop_x, crop_y)

PR #420, opened 2021-09-06
replaced baseclass x,y,z,realm members with Pos4d.

turleypol:
Adapted the code, no changes except making the compiler happy


frozenblit (approved):
Looks good. I just commented a few places you might want to check again and a suggestions that might make future changes easier ;)

PR #426, opened 2021-09-12
Regions Pos class refactor

turleypol:


pol-core/pol/regions/miscrgn.cpp
```diff
@@ -124,65 +123,29 @@ void WeatherDef::copy_default_regions()
...
-
-    for ( unsigned int i = 0; i < gridwidth; i++ )
-    {
-      for ( unsigned int j = 0; j < gridheight; j++ )
-      {
-        default_regionrealms[realm][i][j] = regionrealms[realm][i][j];
-      }
-    }
+    Range2d area = Range2d( Pos2d( 0, 0 ), XyToZone( realm->area().se() ), nullptr );
+    for ( const auto& p : area )
```
frozenblit, line 127: The only issue here is that Range2d iterates over x first, right? So memory access for the zones could be optimized. I guess it depends how often this access happens

frozenblit (approved):
Looks good. Just note my comment about the order of iterating over the zones using Range2d. I think in the other branch you inverted the indices [x][y] -> [y][x]

PR #428, opened 2021-09-15
wrappers for realm/uworld methods accepting Posclasses.

turleypol:
switched memoryorder of realm zone

### DIFF
--- a/pol-core/pol/uworld.h
+++ b/pol-core/pol/uworld.h
@@ -97,7 +97,7 @@ inline Zone& getzone( unsigned short x, unsigned short y, Realms::Realm* realm )
   passert( x < realm->width() );
   passert( y < realm->height() );
 
-  return realm->getzone( x >> Plib::WGRID_SHIFT, y >> Plib::WGRID_SHIFT );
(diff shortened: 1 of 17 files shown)


pol-core/pol/uworld.h
```diff
@@ -104,7 +97,18 @@ inline Zone& getzone( unsigned short x, unsigned short y, Realms::Realm* realm )
...
+  return realm->getzone( x >> Plib::WGRID_SHIFT, y >> Plib::WGRID_SHIFT );
+}
+
+inline Pos2d zone_convert( const Pos4d& p )
+{
+  return Pos2d( static_cast<unsigned short>( p.x() >> Plib::WGRID_SHIFT ),
+                static_cast<unsigned short>( p.y() >> Plib::WGRID_SHIFT ) );
+}
+
+inline Zone& getzone( const Pos4d& p )
```
frozenblit, line 109: Just to remember that getzone(x,y) and realm->getzone(x,y) are using different units. The first is world location, the second is grid coordinate. Maybe we can think of a better name to distinguish them

frozenblit (approved):
Looks fine, just a small comment about the naming of getzone that might lead to confusion

PR #431, opened 2021-09-28
Improvements on the improvements for uoconvert

frozenblit:
Fixes a few issues with reading UOP files, estimating parameters of tiledata.mul and map dimensions. Also refactors the code a bit. 

Please look at the individual commits for details.


turleypol (approved):
I think it looks good. Only comment I have currently is that the code in uoconvert.cpp is in the namespace Pol::UoConvert and has at the beginning using namespace Plib. So Pol::Plib::MUL::Map::valid_size is a bit long ;) the later using namesplace Plib can thus be also removed.

PR #433, opened 2021-10-08
Change to the references to /pkg/std and /pkg/opt.

Admin-Yukiko:
Since the addition of /pkg/optional for optional packages, the old references are no longer relevant and need to be updated.


turleypol:
Old outdated PR. #434 merged instead.

PR #434, opened 2021-10-09
Updating the references where standard and optional packages might be…

Admin-Yukiko:
… placed.

Since adding /pkg/optional as the new location for optional packages, the eScript guide needed updating.


sigmantium (approved):
Yep looks good to me.

PR #443, opened 2021-11-04
Change outgoing packet handlers to write CP-1252 when appropriate

brndd:
I went through all uses of `PacketWriter::Write( const char* x, u16 len, bool nullterm = true )` and converted the written string to CP-1252 in the packets which should use it. This should fix most encoding errors like this:

broken1

So that they appear correctly like this:
broken2

Characters not in the CP-1252 code page will be replaced by question marks.

My methodology for the conversion was this:

- If the packet has a fixed-length string, simply convert by wrapping the string in the `strUtf8ToCp1252()` function in the `Write()` call. This is safe, because CP-1252 strings are always the same length or shorter than UTF-8 strings.
- If the packet has a null-terminated string, the conversion is done a little bit before the `Write()` call, depending a little on the exact implementation of the function, so that the function gets the converted string's (rather than the original UTF-8 string's) length for the various checks that are usually done.

-----

~~I also updated `Clib::sanitizeUnicodeWithIso()` so that it "sanitizes" with CP-1252 instead. This function seems quite poorly named; it is apparently mostly used in some BScript stuff and things like speech packet handlers to convert CP-1252 encoded text into UTF-8. It currently does not do that (correctly, at least), and I have written a test which demonstrates this problem.~~ This function seems to be intended to be used for text from script and config files. It's not actually used for anything too important on incoming packets, so it's not relevant.

### DIFF
--- a/pol-core/pol/testing/testmisc.cpp
+++ b/pol-core/pol/testing/testmisc.cpp
@@ -225,6 +225,38 @@ void test_convertquotedstring()
   test_dqs( "\" \\\"hi\"", " \"hi" );
 }
 
+void test_sanitizeUnicodeWithIso()
+{
+  std::string input;
+  std::string output;
+  std::string expected;
+
+  input = "Some weird characters: \x87 \x95 \x80.";
+  output = std::string(input);
+  expected = "Some weird characters: ‡ • €.";
+  Clib::sanitizeUnicodeWithIso(&output);
+  if ( output != expected )
+  {
+    INFO_PRINT << "sanitizeUnicodeWithIso fails!\n\tinput: " << input << "\n\toutput: " << output << "\n\texpected: " << expected << "\n";
+    UnitTest::inc_failures();
+  }
+  else
+    UnitTest::inc_successes();
+
+  input = "Maybe someone just wants to say \xC3\xA4 in CP-1252.";
+  output = std::string(input);
+  expected = "Maybe someone just wants to say Ã¤ in CP-1252.";
+  Clib::sanitizeUnicodeWithIso(&output);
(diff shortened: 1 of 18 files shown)


pol-core/pol/testing/testmisc.cpp
```diff
@@ -225,6 +225,38 @@ void test_convertquotedstring()
...
+  output = std::string(input);
+  expected = "Maybe someone just wants to say Ã¤ in CP-1252.";
+  Clib::sanitizeUnicodeWithIso(&output);
+  if ( output != expected )
+  {
+    INFO_PRINT << "sanitizeUnicodeWithIso fails!\n\tinput:    " << input << "\n\toutput:   " << output << "\n\texpected: " << expected << "\n";
+    UnitTest::inc_failures();
+  }
+  else
+    UnitTest::inc_successes();
```
brndd, line 256: This is the test that demonstrates the problem with the current implementation of `sanitizeUnicodeWithIso()`. It is impossible for any CP-1252 text passing through this function to contain the character combination `Ã¤` (and many other combinations that are valid UTF-8 surrogates).

brndd:
It seems incoming packet handlers are already fine, at least as far as encoding is concerned. We could maybe do with some sanitization on them; currently speech packets are simply relayed as they come in from the client, so some prankster might insert control characters or other weird stuff in the text that could result in unexpected behaviour, but that's not an encoding-related issue.

turleypol (approved):
Looking good, a small corechanges entry is only missing.

brndd:
> Looking good, a small corechanges entry is only missing.

Fixed!

PR #446, opened 2021-12-10
Fixed move of mobiles/items between realms

turleypol:
Broken due to pos rewrite.
Multis will still crash

Added second realm to the testsuite to be able to test it.
Fixed login seed pkt for testclient.
The core no more sends new subserver pkt when moving a client in the same realm.


frozenblit (approved):
Looks good :D

PR #448, opened 2021-12-27
use std::filesystem to iterate over directories

turleypol:
realms folder can now contain files


frozenblit (approved):
Looks good. Maybe the only remaining question is why you changed files from std::vector<> to std::set<> in ecompilemain? :p

PR #455, opened 2022-02-17
Fixed reading map properties + VS2022 boost update

AsYlum-:
Small bug fix after last coordinates refactoring.


frozenblit (approved):
Thanks!

PR #462, opened 2022-06-15
Fix crash on unknown member access on Client; add Client.acct

KevinEady:


frozenblit (commented):
Good catch! Shouldn't there also be a core-changes entry? ;)

KevinEady:
> Good catch! Shouldn't there also be a core-changes entry? ;)

Addressed @frozenblit

turleypol (approved):

PR #475, opened 2022-09-27
Add force parameter to CreateNPCFromTemplate()

brndd:
There are times where I want to create NPCs in unwalkable locations, such as inside of walls, or in water, or create sea creatures on land. I worked around this in EScript by creating the NPC outside of the map and then force-moving it to its final destination, but I figured it's silly to have to use such workarounds instead of just updating the core. 

This patch seems to work, but I must confess I'm not terribly familiar with the core's workings so there is a chance that I messed something up.

Another option I considered was adding a flag variable to allow for more granular control of exactly what conditions are ignored, but I couldn't think of enough flags that would really be useful. It would've been something like "ignore doors", "ignore movemode", "ignore everything". If that sounds like a better solution to you, I am open to changing this PR to do that instead.

### DIFF
--- a/pol-core/support/scripts/uo.em
+++ b/pol-core/support/scripts/uo.em
@@ -274,7 +274,7 @@ CreateItemInContainer( container, objtype, amount := 1 );
 CreateItemInInventory( container, objtype, amount := 1 );
 CreateMenu( title );
 CreateMultiAtLocation( x, y, z, objtype, flags := 0, realm := _DEFAULT_REALM );
-CreateNpcFromTemplate( template, x, y, z, override_properties := 0, realm := _DEFAULT_REALM);
+CreateNpcFromTemplate( template, x, y, z, override_properties := 0, realm := _DEFAULT_REALM, force := 0);
 DestroyItem( item );
 DestroyMulti( multi );
 Detach();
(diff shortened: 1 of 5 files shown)


pol-core/support/scripts/uo.em
```diff
@@ -274,7 +274,7 @@ CreateItemInContainer( container, objtype, amount := 1 );
 CreateItemInInventory( container, objtype, amount := 1 );
 CreateMenu( title );
 CreateMultiAtLocation( x, y, z, objtype, flags := 0, realm := _DEFAULT_REALM );
-CreateNpcFromTemplate( template, x, y, z, override_properties := 0, realm := _DEFAULT_REALM);
+CreateNpcFromTemplate( template, x, y, z, override_properties := 0, realm := _DEFAULT_REALM, force := 0);
```
turleypol, line 277: I would suggest to use a more verbose name for this parameter.
"force" could be anything:)
Maybe something like "forcelocation" that it has equal naming like the flag for MoveObject.

turleypol (commented):
Rest is fine, but I think its not the best parameter name.

pol-core/support/scripts/uo.em
brndd: Fixed. Sorry it took a while, been a busy couple of weeks.

PR #476, opened 2022-09-27
Add new EScript function: GetStandingCoordinates, and a new option to GetStandingLayers

brndd:
Adds a new EScript function with the following signature:
`GetStandingCoordinates( x, y, radius := 0, minz := -128, maxz := 127, realm := _DEFAULT_REALM, movemode := "L", doors_block := 0)`

This function takes in an x/y/realm coordinate and a movemode, and returns a list of valid locations to stand on within a given radius of the coordinates. The `minz` and `maxz` parameters can be used to limit the z-range of the search. The `movemode` parameter defines the movemode that's used to decide whether a tile can be stood on. `doors_block` excludes tiles with doors on them from the search.

One example use-case of this function is to get a list of valid locations to spawn an NPC in. Another example could be to find a valid location around a player for an enemy to pathfind to.

Example output from a debug script when targeting the tile under the cursor:

2023-02-02T21:33:48

-----

I also added a new parameter, `includeitems`, to a similar pre-existing function named `GetStandingLayers`. This function can essentially be used to return a list of all collidable items in a tile. It's pretty confusing, though, and honestly not very useful. I found a small bug in this function too, which I fixed: the height of the returned tiles was hardcoded to 1, instead of the actual height of the tile.

<details>
<summary>Old description</summary>
This is a weird EScript function that basically returns the height, standable z-level (or z-coordinate, if the object cannot be stood on) and list of matched flags for every static, multi and now item at a coordinate that matches the mapdata flags given.

I wanted to use it to get every standable position at a given x, y coordinate, but it didn't work for that because it did not consider items, so I added that functionality to it.

However, as I was doing this I started questioning the existence of this function. The name is bad, as it does way more than the name suggests. The returned data is unintuitive: if you can stand on the object, the function returns the z-coordinate of its top (i.e. the z-coordinate a character stood on top would be located at), but otherwise it returns the z-coordinate of its bottom.

I also noticed some weirdness with the data the function returns. On tile A, where the shoreline drops off towards Z-15, it says the wooden surface's height is 20. On tile B the height is correctly reported as 5.

### Tile A
2022-09-28T01:21:24

### Tile B
2022-09-28T01:23:45

There was also a bug affecting the z-coordinate returned, as the function would add 1 to the z-coordinate of walkable tiles, rather than their height. This resulted in nonsense results for walkable tiles with a height different from 1.

----

Perhaps it would be better to deprecate this function and create a new one with a better name that does not feature as many idiosyncrasies. Is that something we can do?
</details>


brndd:
Plan for this PR once I get around to it: implement a new function called `GetStandingCoordinates(x, y, radius := 0, zmin := ZCOORD_MIN, zmax := ZCOORD_MAX, realm := britannia, movemode := "L")` that returns an array of valid standing coordinates (x, y, z, realm). Works in a radius, but the radius defaults to 0 so by default it only returns info for a single tile.

brndd:
Took me a while to get around to this, and then actually implementing it took quite a bit longer than I expected too.

pol-core/pol/module/uomod.cpp
```diff
@@ -4471,6 +4477,68 @@ BObjectImp* UOExecutorModule::mf_GetStandingLayers( /* x, y, flags, realm */ )
...
 
+BObjectImp* UOExecutorModule::mf_GetStandingCoordinates(/* x, y, radius, minz, maxz, realm := _DEFAULT_REALM, movemode := "L", doors_block = 0 */)
+{
+  int dummy, x, y, r, minz, maxz;
+  const String* realmname;
+  const String* movemodename;
+  bool doors_block;
+
+  if (!(
+           getParam(0, dummy) && //FIXME: remove this dummy once you figure out why getting the 0th param fails every time
```
brndd, line 4488: There is a bizarre bug here that I could not track down. The first param given to this function would always have the value `0`. No matter what is passed as the value here, `dummy` will always be `0`. This will obviously need to be fixed before merging, but I have no idea what's going wrong :-P

pol-core/pol/module/uomod.cpp
```diff
@@ -4471,6 +4477,69 @@ BObjectImp* UOExecutorModule::mf_GetStandingLayers( /* x, y, flags, realm */ )
...
+UOExecutorModule::mf_GetStandingCoordinates() /* x, y, radius, minz, maxz, realm := _DEFAULT_REALM,
+                                                 movemode := "L", doors_block = 0 */
+{
+  int dummy, x, y, r, minz, maxz;
+  const String* realmname;
+  const String* movemodename;
+  bool doors_block;
+
+  if ( !( getParam( 0, dummy ) &&  // FIXME: remove this dummy once you figure out why
+                                             // getting the 0th param fails every time
```
brndd, line 4490: There is a bizarre bug here that I could not track down. The first param given to this function would always have the value `0`. No matter what is passed as the value here, `dummy` will always be `0`. This will obviously need to be fixed before merging, but I have no idea what's going wrong :-P

brndd: This was because of the `reinterpret_cast` of bool to int some lines down, as pointed out by Turley.

pol-core/pol/realms/realm.h
```diff
@@ -214,6 +226,28 @@ class Realm
   }
   void readmultis( Plib::StaticList& vec, const Core::Pos2d& pos ) const;
 
+  void readdynamics( Plib::MapShapeList& vec, const Core::Pos2d& pos,
+                     Core::ItemsVector& walkon_items, bool doors_block, unsigned int flags ) const;
+
+  void readdynamics( Plib::MapShapeList& vec, unsigned short x, unsigned short y,
+                     Core::ItemsVector& walkon_items, bool doors_block, unsigned int flags ) const
+  {
+    return readdynamics( vec, Core::Pos2d( x, y ), walkon_items, doors_block, flags );
```
turleypol, line 235: Same here

pol-core/pol/realms/realm.h
```diff
@@ -214,6 +226,28 @@ class Realm
...
+                     Core::ItemsVector& walkon_items, bool doors_block ) const
+  {
+    return readdynamics( vec, Core::Pos2d( x, y ), walkon_items, doors_block,
+                         Plib::FLAG::WALKBLOCK );
+  }
+
+  void readdynamics( Plib::MapShapeList& vec, const Core::Pos2d& pos,
+                     Core::ItemsVector& walkon_items, bool doors_block ) const
+  {
+    return readdynamics( vec, pos, walkon_items, doors_block, Plib::FLAG::WALKBLOCK );
```
turleypol, line 248: I would give the function a different name, it's unclear only due to slightly different arguments that WALKBLOCK is here used

brndd:
Todo: write a test for this.

brndd:
Well, that took me a little longer than it should have, but there's now a test. It's a very simple one because I couldn't figure out how to actually run the test map in a regular client, so I just logged some results from the test map and used those as baseline good data to compare against.

The function signature in uo.em also had some outdated nonsense in it, so I fixed that. Marked all the review conversations as resolved, as the issues in them were addressed.

I'll do a little bit more testing by hand tomorrow, but other than that this should now be ready as far as I'm concerned.

brndd:
Turns out it was just a minor bug. Now this should be ready.

KevinEady (approved):
LGTM

PR #477, opened 2022-10-09
Method function crash

turleypol:
Fixed crash when calling another method inside of a method script.
Fixed a shutdown memory leak in the unittest
Added tests for the crash


KevinEady (approved):
LGTM!

PR #480, opened 2022-10-22
Update unicode.em and uo.em to fix casing with Broadcast* requiredCmdLevel

che666:
Update unicode.em and uo.em to fix casing with Broadcast* requiredCmdLevel

Please see:
https://docs.polserver.com/pol100/fullfunc.php?xmlfile=uoem#Broadcast

and:
https://docs.polserver.com/pol100/fullfunc.php?xmlfile=unicodeem#BroadcastUC

### DIFF
--- a/pol-core/support/scripts/unicode.em
+++ b/pol-core/support/scripts/unicode.em
@@ -28,7 +28,7 @@ const _DEFAULT_UCTEXT_REQUIREDCMD := 0;
 // 'langcode' is a 3-character "Originating Language" code
 //       e.g. ENG, ENU, CHT, DEU, FRA, JPN, RUS, KOR (etc??)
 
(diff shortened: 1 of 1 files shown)


che666:
Please merge together with 481

turleypol (changes requested):
When changing the parameter which results in a compilation error for users which have it already in there codebase I would suggest to use underscores as the other parameters to make it more consistent.
So eg `required_cmdlevel`
The docs (docs/docs.polserver.com/pol100/uoem.xml) need to be also adapted, it's currently undecided between CamelCase and LowerCase.
Since it will break compilation a core-change entry needs to be added: pol-core/doc/core-changes.txt, afterwards execute doctools/buildcorechanges.py to generate the webversion of the corechanges.

che666:
Will do all of the above. Please allow me time to work on it.

turleypol:
Will be fixed with #481

PR #482, opened 2022-10-23
Client related threading issues

turleypol:
busy flag for DynTaskPool needs to be atomic 
client SocketMutex is no more static
replaced Client::Delete with a deconstructor
Client::disconnect needs to be atomic
Multiple threads want to know UOClientListener::login_clients size.
CI action which runs with Thread/AddressSanitizer
updated some github actions

### DIFF
--- a/pol-core/pol/network/client.cpp
+++ b/pol-core/pol/network/client.cpp
@@ -69,7 +69,6 @@ void cancel_trade( Mobile::Character* chr1 );
 namespace Network
 {
 unsigned int Client::instance_counter_;
-std::mutex ThreadedClient::_SocketMutex;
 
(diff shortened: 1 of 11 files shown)


pol-core/pol/network/client.cpp
```diff
@@ -134,17 +133,12 @@ Client::Client( ClientInterface& aInterface, Crypt::TCryptInfo& encryption )
   memset( &versiondetail_, 0, sizeof( versiondetail_ ) );
 }
 
-void Client::Delete( Client* client )
```
frozenblit, line 137: ❤

frozenblit (approved):

KevinEady (approved):
github yamls LGTM

Issue #487, opened 2022-12-11
Feature Request for CreateItemInBackpack, CreateItemInContainer, and CreateItemInInventory.


Admin-Yukiko:
ThisIsMe requested this in the Discordance server. So I thought I should create an issue so it does not get lost in the text stream.

`MoveItemToContainer( item, container, x := -1, y := -1, add_to_existing_stack := 0 )` has an x and y position for placing an item at specific coordinates within a container.
The request is to add the same optional parameters to ` CreateItemInBackpack, CreateItemInContainer`, and `CreateItemInInventory.`
```
CreateItemInBackpack( of_character, objtype, amount := 1, x:=-1, y:=-1 )
CreateItemInContainer( container, objtype, amount := 1, x:=-1, y:=-1 )
CreateItemInInventory( container, objtype, amount := 1, x:=-1, y:=-1 )
```

PR #490, opened 2022-12-13
Add x, y parameters to create item function calls

KevinEady:
Adds `x` and `y` optional parameters to `CreateItemInBackpack`, `CreateItemInContainer`, and `CreateItemInInventory` functions.

Closes: #487


turleypol (approved):

pol-core/pol/module/uomod2.cpp
```diff
@@ -779,8 +779,8 @@ BObjectImp* UOExecutorModule::mf_SendSellWindow( /* character, vendor, i1, i2, i
 }
 
 extern BObjectImp* _create_item_in_container( UContainer* cont, const ItemDesc* descriptor,
-                                              unsigned short amount, bool force_stacking,
-                                              UOExecutorModule* uoemod );
+                                              unsigned short amount, bool force_stacking, short px,
```
turleypol, line 782: optional<Pos2d> would be better I think, but its no showstopper

PR #492, opened 2022-12-17
Potential login crash

turleypol:
client_io_thread returns false if the client will be deleted.
Since we didn't checked here the return value we have a dangling pointer if the thread who deletes the client is faster then the later loop which correctly checks the return value.


KevinEady (approved):
LGTM!

PR #498, opened 2023-01-06
reworked get_item and its undo behaviour

### LINKED ISSUE #488: NoDrop & NoDropException fun
When setting an item to no_drop and a container to no_drop_exception OTHER than my players pack, when I place the item in the world, it bounces it to my backpack which is set to no_drop_exception instead of the container from whence it came.

If I then in the CanInsertScript of my backpack disallow the item from being able to add it and repeat moving the item from it's original container to the world, instead of going into my backpack it goes on the ground at my character's feet.

If I then in the CanInsertScript force move the item and disallow it from being added to my backpack, the shard crashes with the following error:
`remove_item_from_world: item 0x598293C1 at 0,0 does not exist in world zone ( Old Serial: 0x598293C1 )
Assertion Failed: itr != zone.items.end(), D:\GitHub\polserver\pol-core\pol\uworld.cpp, line 62`

I do not receive any of the typical shard crash files but am presented with the above error in the console.

If an item is set to NoDrop and the following conditions occur:
Dropped in the World
Dropped in a container NOT set to exceptioned
Rejected by CanInsert for various reasons
Dropped onto an NPC NOT set to exceptioned

I would think (and like) the correct reaction would or should be the following:
Returned to the container from whence it came preferably at the X and Y coordinates in that container it was at when it was first picked up.

The zip file attached should be able to be compiled in the Modern Distro. If you modify the CanInsert script tied to character's backpack with the following simple check you can recreate the crash completely.
`if( adding_item.IsGamePiece() )
	adding_item.ReturnGamePiece( mobile );
	return 0;
endif`
[games.zip](https://github.com/polserver/polserver/files/10205519/games.zip)

turleypol:
old position gets stored and in case of failures restored.
Should fix #488


KevinEady:
Closed in favor of #514

PR #505, opened 2023-02-20
Fix crash related to item position in multi houses

AsYlum-:
This one is for multi -> house as it looks like it was missed when fixing the crashes last time.

Previous fix: https://github.com/polserver/polserver/commit/6f0db1810a9ebbc21528bdca20bf13375a996401
Original refactoring: https://github.com/polserver/polserver/commit/eeb57d489ace48a973459dafde4cdeac1312ba80


KevinEady (approved):
LGTM! Pumpkins has said this fixes his crashing issue. I noticed that #502 does not have a core-changes, so I suppose it's fine here as well.

PR #507, opened 2023-02-28
Include items on exterior stairs of custom houses in MultiRef.items

KevinEady:
Items on exterior stairs of custom houses were not being included in the multi's `item` script member.

Here is a script where I target the robe, check `robe.multi`, and loop through `robe.multi.items`: 

```
    var item := Target(who).multi;
    Broadcast($"multi: {item}");
    foreach child_item in (item.items)
        Broadcast($"  child item: {child_item.objtype} - {child_item.desc}");
    endforeach
```

First iteration shows the robe not included, and second iteration shows robe included:

Screenshot 2023-02-27 at 22 56 04

### DIFF
--- a/pol-core/pol/multi/customhouses.h
+++ b/pol-core/pol/multi/customhouses.h
@@ -80,6 +80,7 @@ typedef std::list<CUSTOM_HOUSE_ELEMENT> HouseFloorZColumn;
 // v v v v v v
 // [][][][][][] - lists of zoffsets
 // [][][][][][]
+typedef std::list<Items::Item*> ItemList;
 
 class CustomHouseElements
 {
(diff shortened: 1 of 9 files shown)


pol-core/pol/multi/customhouses.h
```diff
@@ -175,6 +176,8 @@ void CustomHousesSendFull( UHouse* house, Network::Client* client,
 void CustomHousesSendFullToInRange( UHouse* house, int design, int range );
 void CustomHousesSendShort( UHouse* house, Network::Client* client );
 void CustomHouseStopEditing( Mobile::Character* chr, UHouse* house, bool send_pkts = true );
+void CustomHouseStopEditing( Mobile::Character* chr, UHouse* house, ItemList& foo,
```
turleypol, line 179: foo? :)
A comment why these two versions exists would be good when we cannot get rid of the second version, otherwise we will forget it.

turleypol (approved):

PR #533, opened 2023-06-12
Diagnostics: Disable Optimizer in analysis mode

KevinEady:
The optimizer will replace constants _inside_ the AST. The AST must be un-optimized in order to properly find constant references.


KevinEady:
This _really_ won't work; a lot of downhill things (eg. `case` statements) require things to be constant/optimizable.

PR #535, opened 2023-08-25
Change packet used for status updates for own char

KevinEady:


pol-core/pol/ufunc.cpp
```diff
@@ -1883,7 +1883,7 @@ void send_create_mobile_to_nearby_cansee( const Character* chr )
                                                     } );
 }
 
-void send_move_mobile_to_nearby_cansee( const Character* chr )
+void send_move_mobile_to_nearby_cansee( const Character* chr, bool include_self )
 {
   MoveChrPkt msgmove( chr );
```
frozenblit, line 1888: I wonder if we have similar updates in other places. Could it make sense to have a facade coordinating the different packets and simplify the code here?

frozenblit (approved):
looks good ;)

PR #547, opened 2023-09-23
Add ServSpecOpt UndoGetItemEnableRangeCheck and UndoGetItemDropHere to control get item undo behaviour

murderinc:
"UndoGetItemEnableRangeCheck" 1/0 default 0.
If enabled - on fail to equip item back on character (if it has been taken from character layer) and on fail to put item back to backpack will check if item has been taken from position/container that is in default_accessible_range. If origin position/container is not in default_accessible_range will skip try to put item to origin position/container (unless item is no_drop) and as last resort will drop item at character position (feet). UndoGetItemDropHere executed before UndoGetItemEnableRangeCheck if both are set.

"UndoGetItemDropHere" 1/0 default 0.
If enabled - on fail to equip item back on character (if it has been taken from character layer) and on fail to put item back to backpack drops item at character position (feet), except no_drop items. Attempt to return item in origin container will be skipped.


testsuite/pol/config/servspecopt.cfg
turleypol, line 1: Why have you added an empty file?

turleypol (changes requested):
Only roughly checked the logic, @KevinEady you where more involved into the discussion

turleypol (approved):

pol-core/pol/getitem.cpp
```diff
@@ -291,28 +292,50 @@ void GottenItem::undo( Mobile::Character* chr )
...
+         && ( !Core::settingsManager.ssopt.undo_get_item_drop_here || _item->no_drop() ) )
     {
-      // Attempt to place the item in the player's backpack.
+      auto* orig_obj = system_find_object( _cnt_serial );
+      if ( orig_obj && orig_obj->isa( UOBJ_CLASS::CLASS_CONTAINER ) )
+      {
+        if ( _item->no_drop()
+             || chr->can_moveanydist()
+             || !Core::settingsManager.ssopt.undo_get_item_enable_range_check
+             || chr->pos().pol_distance( _pos )
```
KevinEady, line 315: As discussed in Discord, should check the distance to `orig_obj->toplevel_pos()` and not `_pos`, as this is the position inside the gump container.

KevinEady (changes requested):

KevinEady (approved):
LGTM! Thanks for spending the time on this! 👍

PR #550, opened 2023-10-14
fixed ReloadConfiguration()

turleypol:
Datarace of client settings
Memleak of NPC templates
Increased the timeout of shard test, 2minutes is especially for the sanitizer builds now not enough.


KevinEady (approved):
LGTM. I applied this commit onto #546 and sanitizer build passes now https://github.com/polserver/polserver/actions/runs/6526228926

PR #554, opened 2023-10-19
rewrite of FindPath

turleypol:
Uses now position/range classes
Extended the params to not only have the blocking spots but all common things per node to minimize the memory per node.
General modernization 
added parameter for movemode ("LSA") before it was only possible to use findpath on land
Testscript added.


KevinEady (approved):
LGTM! 🎆

PR #555, opened 2023-10-23
use nondeprecated openssl api

turleypol:
cleanup refptr
added test for md5_encrypt


pol-core/clib/clib_MD5.cpp
```diff
@@ -88,24 +83,29 @@ void MD5_Cleanup()
...
 namespace Pol
 {
 namespace Clib
 {
 bool MD5_Encrypt( const std::string& in, std::string& out )
 {
-  unsigned char sum[16];
+  auto ctx = EVP_MD_CTX_new();
+
+  EVP_DigestInit_ex( ctx, EVP_md5(), nullptr );
```
frozenblit, line 96: shouldn't one check the return values?

PR #556, opened 2023-10-25
removed pol_distance/inrange* functions

turleypol:
Multis have now a update range based on its footprint

replaced lastxyz with Pos4d. this fixes an old problem that lastxy was always set to 0 on login or realm change and thus all functions like send_if_newly_in_range didn't send anything when the chr is near 0,0.


pol-core/pol/base/position.cpp
```diff
@@ -161,6 +156,12 @@ Pos2d Pos2d::max( const Pos2d& v ) const
   return Pos2d( std::max( _x, v._x ), std::max( _y, v._y ) );
 }
 
+u16 Pos2d::clip_u16( int v )
```
frozenblit, line 159: isn't this just a helper function? why is it in the pos2d class? :D

pol-core/pol/bowsalut.cpp
```diff
@@ -186,62 +187,65 @@ void send_action_to_inrange( const Mobile::Character* obj, UACTION action,
...
-      oldanim.backward = static_cast<u8>( backward );
-      oldanim.repeatflag = static_cast<u8>( repeatflag );
-      oldanim.delay = delay;
-      MobileTranslate::NewAnimDef newanim;
-      bool supports_mount = true;
-      build = true;
-      MobileTranslate const* translate = nullptr;
-      for ( const auto& translates : Core::gamestate.animation_translates )
+  WorldIterator<OnlinePlayerFilter>::InRange(
+      obj, obj->update_range(),
```
frozenblit, line 191: maybe a dumb question: is the update_range() here how much the client wants to see (like the client setting) or how big is their "effect" range?

pol-core/pol/base/position.cpp
turleypol: Good point, working on it :)

pol-core/pol/bowsalut.cpp
turleypol: And bug found.. thx :)
obj->update_range() is the visibility distance so for chars their line of sight and for objects it could be higher due to multi footprint

pol-core/pol/base/position.cpp
turleypol: Done

pol-core/pol/item/weapon.cpp
```diff
@@ -458,7 +458,7 @@ bool UWeapon::consume_projectile( Core::UContainer* cont ) const
 
 bool UWeapon::in_range( const Mobile::Character* wielder, const Mobile::Character* target ) const
 {
-  unsigned short dist = pol_distance( wielder, target );
+  unsigned short dist = wielder->pos().pol_distance( target->pos() );
```
frozenblit, line 461: what do you think about `wielder->distance_to(...)`?

pol-core/pol/mobile/charactr.cpp
```diff
@@ -3638,12 +3631,14 @@ void Character::unhide()
...
       send_owncreate( client, this );
 
-    Core::WorldIterator<Core::OnlinePlayerFilter>::InVisualRange(
+    Core::WorldIterator<Core::OnlinePlayerFilter>::InMaxVisualRange(
         this,
         [&]( Character* chr )
         {
           if ( chr == this )
             return;
+          if ( !chr->in_visual_range( this ) )
```
frozenblit, line 3640: just out of curiosity: why isn't in_visual_range() within is_visible_to_me()? optimization?

frozenblit (approved):

pol-core/pol/pol.cpp
```diff
@@ -489,9 +494,9 @@ void handle_resync_request( Network::Client* client, PKTBI_22_SYNC* /*msg*/ )
   client->send_pause();  // dave removed force=true 5/10/3, let uoclient.cfg option determine xflow
                          // packets (else this hangs 4.0.0e clients)
 
-  Core::WorldIterator<Core::MobileFilter>::InVisualRange(
-      client->chr,
-      [&]( Mobile::Character* zonechr ) { send_client_char_data( zonechr, client ); } );
+  Core::WorldIterator<Core::MobileFilter>::InRange( client->chr->pos(), client->chr->los_size(),
```
KevinEady, line 497: I think this should be a MaxVisualRange + in_range? I believe this should eg. show large multis.

turleypol: It's a MobileFilter so I think InMaxVisualRange is not needed, we need to know here all mobiles which are in range of the current char. The visual size of all characters is 1

KevinEady:
From looking through the changes, these are the things that were touched/modified:
```
send action to inrange

double click

drop item: on mobile, ground, object

pickup item

stat request via packet 34 (range decreased from 20 to 18)

inform item: leftarea, enteredarea, moved

query party / guild locations

handle allnames (range decreased from 20 to 18)

attacking: entering combat (range decreased from 20 to 18)

ghost: hiding ghost when out of war, creating ghost on death

moving characters, war mode change, Character::mightsee, 

walkon script params 

npc: anchors, pc spoke, inform (leftarea, enteredarea, moved)

changing privs to now see/unsee: ghosts, hidden, items, 

module functions: CreateNpcFromTemplate, PlayLightningBoltEffect, List*

movement: send_char_if_newly_inrange, send_item_if_newly_inrange, send_multi_if_newly_inrange, send_objects_newly_inrange_on_boat, remove_objects_inrange, handle_walk to close secure trade window when range > 3

boats: send_smooth_move_to_inrange, send_display_boat_to_inrange, move_travellers, turn_travelers, move_xy, move, move_components, destroy_boat, 

custom houses: send to in range

house: walkon script, script method add/erase_house_part

broadcast client vitals changed

packet: script method sendareapacket 

party: send stat to party

pol: send inrange items, send inrange multis, client resync request (via packet 22)

realm: notified moved, hidden, entered, left 

repsys: propagate criminal to 32 in range

sngclick: single click

speech: to clients, to npcs

textcmd: .resendchars

tooltips: send_object_cache_to_inrange

ufunc: send_remove_character_to_nearby/cansee/cantsee, send_remove_object_to_inrange, send_corpse_equip_inrange, send_item_to_inrange, effects (sounds, moving), death message, move item, send_multi_to_inrange, send_create_mobile_*

UWeapon::in_range
```

pol-core/pol/mobile/charactr.cpp
turleypol: Just checked: for most of the places it would be simpler to add it to. Then there are lines where a new position needs to be checked, ok an additional param with newpos. But then there is something like eg speech this needs to check if visible but with a different range... Gna
So sadly not a simple codecleanup

pol-core/pol/item/weapon.cpp
turleypol: #581

pol-core/pol/mobile/charactr.cpp
turleypol: #581

pol-core/pol/irequest.cpp
```diff
@@ -157,7 +157,7 @@ void srequest( Network::Client* client, PKTIN_34* msg )
         return;
       if ( !client->chr->is_concealed_from_me( bob ) && client->chr->is_visible_to_me( bob ) )
       {
-        if ( pol_distance( client->chr->x(), client->chr->y(), bob->x(), bob->y() ) < 20 )
+        if ( client->chr->in_visual_range( bob ) )
```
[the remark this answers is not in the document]
turleypol: #581

frozenblit (approved):

PR #557, opened 2023-11-01
Add support to restart Item control scripts via RestartScript

KevinEady:


pol-core/pol/module/uomod.cpp
```diff
@@ -3734,9 +3735,14 @@ BObjectImp* UOExecutorModule::mf_RestartScript()
     }
     else
     {
-      return new BError( "RestartScript only operates on NPCs" );
+      return new BError( "RestartScript only operates on NPCs and Items" );
     }
   }
+  else if ( getItemParam( 0, item ) )
```
turleypol, line 3741: A failed getXParam sets func_result_ with an Error, this error will then be returned independent of what this function returns.
You need to use getUObjectParam similar to what eg mf_MoveObjectToLocation uses and check via script_isa

turleypol (approved):

PR #561, opened 2023-11-08
Uoconvert improvements 3

frozenblit:


pol-core/plib/mul/map.h
```diff
@@ -0,0 +1,138 @@
...
+
+    if ( height % blockHeight != 0 )
+      return false;
+
+    size_t nblocks = filesize / Map::blockSize;
+    return expected_blocks( width, height ) == nblocks;
+  }
+};
+static_assert( Map::blockSize == 196, "Size mismatch" );
+static_assert( Map::valid_size( 77070336, 6144, 4096 ) );                 // Legacy map
```
turleypol, line 34: Maybe move these checks to the unittest and not as static_assert? blocksize check looks like something a platform could change due to type sizes so ok to test it as static_assert. But something like valid_size() is more a unittest.

pol-core/plib/uofile00.cpp
```diff
@@ -43,10 +119,21 @@ bool open_uopmap_file( const int mapid )
   }
 
   uopmapfile.open( filename, std::ios::binary );
-  return (bool)uopmapfile;
+
+  if ( (bool)uopmapfile )
```
turleypol, line 123: if (uopmapfile.good()) or !uopmapfile.fail() would be clearer to read, so you don't have to check the type of the variable to check if the standard defines a operator bool ;)

pol-core/uoconvert/UoConvertMain.cpp
```diff
@@ -1052,42 +1053,31 @@ int UoConvertMain::main()
   else if ( command == "map" )
   {
     UoConvert::uo_mapid = programArgsFindEquals( "mapid=", 0, false );
-    UoConvert::uo_usedif = programArgsFindEquals( "usedif=", 0, false );
+    UoConvert::uo_usedif = programArgsFindEquals( "usedif=", 1, false );
```
turleypol, line 1056: Why did you switched the default?

pol-core/pol/polcfg.cpp
```diff
@@ -78,12 +78,10 @@ void PolConfig::read_pol_config( bool initial_load )
...
 
-    unsigned short max_tile = elem.remove_ushort( "MaxTileID", UOBJ_DEFAULT_MAX );
-
-    if ( max_tile != UOBJ_DEFAULT_MAX && max_tile != UOBJ_SA_MAX && max_tile != UOBJ_HSA_MAX )
-      Plib::systemstate.config.max_tile_id = UOBJ_DEFAULT_MAX;
-    else
-      Plib::systemstate.config.max_tile_id = max_tile;
+    if ( elem.has_prop( "MaxTileID" ) )
+    {
+      INFO_PRINT << "Warning: MaxTileID has been removed from pol.cfg.\n";
```
turleypol, line 83: Testsuite has a pol.cfg which contains the setting and we have a Pol.cfg.example

pol-core/uoconvert/UoConvertMain.cpp
```diff
@@ -1231,6 +1244,11 @@ void UoConvertMain::setup_uoconvert()
   unsigned short max_tile =
       static_cast<unsigned short>( programArgsFindEquals( "maxtileid=", 0x0, true ) );
 
+  if ( max_tile )
+  {
+    INFO_PRINT << "Warning: maxtileid has been removed from uoconvert options.\n";
```
turleypol, line 1249: Used in cmake/core_tests.cmake :)

pol-core/uoconvert/UoConvertMain.cpp
```diff
@@ -1346,7 +1362,7 @@ void UoConvertMain::load_uoconvert_cfg()
       else if ( elem.type_is( "ClientOptions" ) )
       {
         if ( elem.has_prop( "UseNewHSAFormat" ) )
-          UoConvert::cfg_use_new_hsa_format = elem.remove_bool( "UseNewHSAFormat" );
+          INFO_PRINT << "Warning: UseNewHSAFormat has been removed from uoconvert.cfg.\n";
```
turleypol, line 1365: Defined in testsuite uoconvert.cfg :)

pol-core/pol/polcfg.cpp
frozenblit: MaxTileID is now only used in POL to further restrict what is contained in tiles.cfg. Docs and configs were updated to represent that

pol-core/uoconvert/UoConvertMain.cpp
frozenblit: Removed there

pol-core/uoconvert/UoConvertMain.cpp
frozenblit: Also removed ;)

pol-core/plib/mul/map.h
frozenblit: I decided to leave those in, though I added other tests. This is just a small sanity check that can be checked at compile time.

KevinEady (approved):
LGTM!

PR #571, opened 2023-12-03
Fix curl libdir on some Linux platforms

brndd:
Curl was failing to link on my Fedora 39 system because it was looking for the built library under `lib/` instead of `lib64/`. This way it should use whatever the platform default is.


KevinEady (approved):
LGTM

PR #574, opened 2023-12-06
Add value parameter validation for PerformAction

Kukkimonsuta:
Using invalid values in `PerformAction` could crash core under some circumstances (win x64, `PerformAction(x, 195)`). This is caused by access to fixed array `old_anim` using passed in value parameter without bounds checks. This PR prevents using out of bounds parameters completely.

During discussion on discord one concern was noted about custom new/clients however in my opinion function `send_action_to_inrange` simply doesn't really support those scenarios properly anyway:

1. it always assumes the action is valid for old animation packet (this is for custom clients with additional actions maybe?)
2. translation buffers are fixed size to `ACTION__HIGHEST` but translation is still attempted so it overflows (source of the potential crash)
3. when `gamestate.animation_translates` is being loaded the map is loaded only up to `ACTION__HIGHEST` so for custom values translations to new packet isn't supported at all

Due to this I have decided to completely prevent passing in values above `ACTION__HIGHEST`. If this is decided to be a problem I'll try to prepare alternative PR that will just avoid crashing (translate only when action is below `ACTION__HIGHEST`) while preserving rest of current behavior.

### DIFF
--- a/docs/docs.polserver.com/pol100/corechanges.xml
+++ b/docs/docs.polserver.com/pol100/corechanges.xml
@@ -2,9 +2,15 @@
 <ESCRIPT>
 	<header>
 		<topic>Latest Core Changes</topic>
-		<datemodified>12-10-2023</datemodified>
+		<datemodified>12-16-2023</datemodified>
 	</header>
 	<version name="POL100.1.0">
+		<entry>
+			<date>12-16-2023</date>
+			<author>Kukkino:</author>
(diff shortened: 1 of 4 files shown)


Kukkimonsuta:
Updated to current master.

We've encountered this crash again today, this time with value 100 meaning it's not consistent on which values it may crash, though that's expected on overflow issues.

frozenblit (approved):

frozenblit:
I wonder why you use such high action values, but the changes look good to me

Kukkimonsuta:
To explain the high values - this shard has very long history and these values are there almost certainly by mistake, we're slowly getting through fixing all usages, but some are harder than others since they can be stored in bunch of cprops as well.

Anyway thank you for accepting, this should help us with stability ❤️

PR #578, opened 2023-12-17
WIP: Increase precision when packing doubles

Kukkimonsuta:
This proof of concept, not intended for merging just yet. This is an attempt to fix very early rounding when packing double numbers, see below. I believe double should be precise to up to 15 digits, currently it's 6.

```
var v := 12345678.0;
SendSysMessage(who, "test set: " + CInt(v));
SetObjProperty(who, "numtest", v);
v := GetObjProperty(who, "numtest");
SendSysMessage(who, "test get: " + CInt(v));
```

image

This means that after 11 days saving value of `ReadMillisecondClock` has only 10 second precision which breaks a lot of usages.

### DIFF
--- a/pol-core/bscript/dbl.cpp
+++ b/pol-core/bscript/dbl.cpp
@@ -5,6 +5,7 @@
 
 
 #include <cmath>
+#include <iomanip>
 #include <sstream>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


Kukkimonsuta:
The way to handle this used in this PR is inappropripate, see https://vitaut.net/posts/2023/printing-double/

Further effort is going on in https://github.com/polserver/polserver/commits/doubletostring , so closing for now

PR #580, opened 2023-12-20
Double to string precision

turleypol:
Uses different std function to convert a double to string without loosing precision.
Minimum MacOS Version is now 13.3
Fixes also cmake output when a escript test fails

To-dos:
* CI build needs a environment variable for minimum version, is this also needed for manual builds? -> Kevin said no problem :)
* verify that we have a testscript which packs and unpacks a double and compare if the number hasn't changed -> added
* core changes -> added


KevinEady (approved):
LGTM!

Issue #581, opened 2023-12-22
Continuation of inrange PR #556


turleypol:
From PR #556
* what do you think about wielder->distance_to(...)? Instead of pos().pol_distance(..)
* just out of curiosity: why isn't in_visual_range() within is_visible_to_me()? optimization?
** Just checked: for most of the places it would be simpler to add it to. Then there are lines where a new position needs to be checked, ok an additional param with newpos. But then there is something like eg speech this needs to check if visible but with a different range... Gna
So sadly not a simple codecleanup
* activate client update range (pkthandler cfg setting for valid range and return the new value, chr::los_size return the client one)
* item/NPC events are limited to 32
   realm/realm.cpp Realm::notify_entered eg
   duplicate the logic? in_visual_range/max_visual_range but for events?
* irequest.cpp: I don't think this check is needed, as statrequest has this check on line 52

frozenblit:
Four of the five bullets here are already marked fixed in the body itself (#583, #726). The one remaining item — item/NPC events limited to 32 — is split out to #931 with the current code references, so closing this.

PR #583, opened 2023-12-26
Codecleanup from #581

turleypol:
is_visible_to_me checks now also in_visual_range
added obj::distance_to
fixed warnings

See #581

### DIFF
--- a/pol-core/pol/irequest.cpp
+++ b/pol-core/pol/irequest.cpp
@@ -38,22 +38,19 @@ namespace Core
 {
 void statrequest( Network::Client* client, u32 serial )
 {
-  if ( serial == client->chr->serial )
+  auto chr = client->chr;
+  if ( serial == chr->serial )
   {
-    send_full_statmsg( client, client->chr );
+    send_full_statmsg( client, chr );
+    return;
   }
-  else
-  {
-    Mobile::Character* chr = find_character( serial );
-    if ( chr != nullptr )
-    {
-      if ( client->chr->is_visible_to_me( chr ) )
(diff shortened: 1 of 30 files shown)


pol-core/pol/irequest.cpp
```diff
@@ -38,22 +38,19 @@ namespace Core
...
+    send_full_statmsg( client, chr );
+    return;
   }
+  Mobile::Character* bob = find_character( serial );
+  if ( !bob )
+    return;
+  if ( chr->is_visible_to_me( bob ) )
+    send_short_statmsg( client, bob );
+  if ( chr->has_party() )
+    chr->party()->send_stat_to( chr, bob );
```
frozenblit, line 53: does this check if bob is in chr's party?

pol-core/pol/irequest.cpp
```diff
@@ -38,22 +38,19 @@ namespace Core
...
-          send_short_statmsg( client, chr );
-      }
-    }
+    send_full_statmsg( client, chr );
+    return;
   }
+  Mobile::Character* bob = find_character( serial );
+  if ( !bob )
+    return;
+  if ( chr->is_visible_to_me( bob ) )
```
frozenblit, line 50: the other code was checking for concealed. Does this code do the same?

pol-core/pol/irequest.cpp
turleypol: Yes is_member( bob->serial) gets called in send_stat_to

frozenblit (approved):

PR #584, opened 2023-12-27
pkt definitions with dynamic entries are now the max possible size

turleypol:
The client holds a buffer which gets implicitly cast to the pkt struct. This means the pktdefs can have the maximum possible size for dynamic entries. Which fixes a compiler warning about reading out of bounds

### DIFF
--- a/pol-core/pol/network/pktboth.h
+++ b/pol-core/pol/network/pktboth.h
@@ -13,6 +13,7 @@
 #define __PKTBOTHH
 
 #include "../clib/rawtypes.h"
+#include "../plib/uconst.h"
 #include "pktbothid.h"
 #include "pktdef.h"
 
@@ -150,9 +151,9 @@ struct PKTBI_3B
     u8 layer;
     u32 item_serial;
     u16 number_bought;
-  } items[1]; /* Dunno how many it should be */
+  } items[( MAXBUFFER - 8 ) / 7];
 };
-static_assert( sizeof( PKTBI_3B ) == 15, "size missmatch" );
+static_assert( sizeof( PKTBI_3B ) == 2556, "size missmatch" );
 
 struct PKTBI_56
 {
@@ -183,9 +184,9 @@ struct PKTBI_66
   u16 pages;
   u16 page;
   u16 lines;
-  char text[1];
+  char text[MAXBUFFER - 13];
 };
-static_assert( sizeof( PKTBI_66 ) == 14, "size missmatch" );
+static_assert( sizeof( PKTBI_66 ) == MAXBUFFER, "size missmatch" );
 
 // struct PKTBI_66_HDR
 //{
@@ -450,9 +451,9 @@ struct PKTBI_9A
   u32 serial;
   u32 prompt;
   u32 type;
-  char text[1];
+  char text[MAXBUFFER - 15];
 };
-static_assert( sizeof( PKTBI_9A ) == 16, "size missmatch" );
+static_assert( sizeof( PKTBI_9A ) == MAXBUFFER, "size missmatch" );
 
 struct PKTBI_B8_IN_REQUEST
 {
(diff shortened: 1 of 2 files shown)


pol-core/pol/network/pktboth.h
```diff
@@ -465,16 +466,17 @@ struct PKTBI_B8_IN_UPDATE
   u32 serial;
   u16 cmdtype;
   u16 textlen;
-  u16 wtext[2];
+  u16 wtext[SPEECH_MAX_LEN + 1];
 };
-static_assert( sizeof( PKTBI_B8_IN_UPDATE ) == 12, "size missmatch" );
+static_assert( sizeof( PKTBI_B8_IN_UPDATE ) == SPEECH_MAX_LEN * 2 + 2 + 8, "size missmatch" );
```
frozenblit, line 471: would it make sense to also check if <= MAXBUFFER?

pol-core/pol/network/pktboth.h
```diff
@@ -150,9 +151,9 @@ struct PKTBI_3B
     u8 layer;
     u32 item_serial;
     u16 number_bought;
-  } items[1]; /* Dunno how many it should be */
+  } items[( MAXBUFFER - 8 ) / 7];
 };
-static_assert( sizeof( PKTBI_3B ) == 15, "size missmatch" );
+static_assert( sizeof( PKTBI_3B ) == 2556, "size missmatch" );
```
frozenblit, line 156: is there a better constant here than this magic number?

pol-core/pol/network/pktboth.h
```diff
@@ -150,9 +151,9 @@ struct PKTBI_3B
     u8 layer;
     u32 item_serial;
     u16 number_bought;
-  } items[1]; /* Dunno how many it should be */
+  } items[( MAXBUFFER - 8 ) / 7];
```
frozenblit, line 154: just to be sure: is maxbuffer also the maximum msglen? I guess otherwise we would have problems already

pol-core/pol/network/pktboth.h
```diff
@@ -678,7 +681,7 @@ struct PKTBI_BF_14
   u16 unk;  // always 00 01
   u32 serial;
   u8 numentries;
-  PKTBI_BF_14_ENTRIES entries[1];
+  PKTBI_BF_14_ENTRIES entries[( MAXBUFFER - 7 ) / 8];
```
frozenblit, line 684: this works, but would max(u8)*8 make more sense?

pol-core/pol/network/pktboth.h
```diff
@@ -954,9 +959,9 @@ struct PKTBI_D6_IN
   struct
   {
     u32 serial;
-  } serials[1];
+  } serials[( MAXBUFFER - 3 ) / 4];
 };
-static_assert( sizeof( PKTBI_D6_IN ) == 7, "size missmatch" );
+static_assert( sizeof( PKTBI_D6_IN ) == 2559, "size missmatch" );
```
frozenblit, line 964: this constant appears quite often. What is it? :D

pol-core/pol/network/pktboth.h
frozenblit: And is this related to the 2559 in other spots?

frozenblit (approved):

pol-core/pol/network/pktboth.h
turleypol: It's smaller then 0xffff, but somewhere during pktreveive checked.

pol-core/pol/network/pktboth.h
turleypol: I guess it never changes, the correct thing would be maybe to have a size of(entries) instead of 8 :)

pol-core/pol/network/pktboth.h
turleypol: No just random :)

PR #585, opened 2023-12-29
AuxConnection transmit race

turleypol:
Since the actual transmit is done by a threadpool to not block the script there was a chance that the messages where send in wrong order. Especially when using the newline version (this called 2 times send). 
Happend on very rare occasions in our CI builds with the connection to the client.

### DIFF
--- a/pol-core/pol/network/auxclient.cpp
+++ b/pol-core/pol/network/auxclient.cpp
@@ -237,6 +237,9 @@ void AuxClientThread::run()
     std::this_thread::sleep_for( std::chrono::seconds( 1 ) );
 
   Core::PolLock lock;
+  // hold also the transmit mutex, the counter syncs but not in a way that the threadsanitizer
+  // recognizes is.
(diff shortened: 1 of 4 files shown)


frozenblit (dismissed):

pol-core/pol/network/auxclient.cpp
```diff
@@ -234,6 +237,9 @@ void AuxClientThread::run()
     std::this_thread::sleep_for( std::chrono::seconds( 1 ) );
 
   Core::PolLock lock;
+  // hold also the transmit mutex, the counter syncs but not in a way that the threadsanitizer
+  // recognizes is.
+  std::unique_lock<std::mutex> transmitlock( _transmit_mutex );
```
frozenblit, line 242: is there any situation where the transmitlock might be used before the pollock?

turleypol: It's never used with PolLock, the actual transmit where the lock is needed runs without our global lock

frozenblit (approved):

PR #587, opened 2023-12-30
Character buffs

turleypol:
it's now possible to set for the first cliloc arguments
* addBuff(int icon_id, int duration, int cliloc_name, int cliloc_descr, string desc_arguments, [string name_arguments])
* optional last parameter
added chr.buffs member to return the current buffs as dictionary (iconid, struct{name_cliloc, desc_cliloc, end_time, name_args, desc_args})

Todo: as last optional param it's a non breaking change, but the order is a bit messy.. breaking change? New method version eg "addBuffEx"? Keep it in that order?

### DIFF
--- a/docs/docs.polserver.com/pol100/corechanges.xml
+++ b/docs/docs.polserver.com/pol100/corechanges.xml
@@ -2,9 +2,17 @@
 <ESCRIPT>
 	<header>
 		<topic>Latest Core Changes</topic>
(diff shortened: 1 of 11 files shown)


KevinEady (approved):
LGTM, I think there's a CUO bug with skipping some characters of the title though. Looks fine in OSI

OSI:
osi

CUO:
cuo

PR #592, opened 2024-01-07
New format lib version

turleypol:
Luckily it can exist in parallel to the old version (only include guard rename needed)
Rewrite INFO_PRINT/INFO_PRINT_TRACE to use the new formatting lib
New formatting lib offers compile time checks for formatting and it's the base of the c++20 formatting addition.


pol-core/pol/item/weapon.cpp
```diff
@@ -459,12 +459,14 @@ bool UWeapon::consume_projectile( Core::UContainer* cont ) const
 bool UWeapon::in_range( const Mobile::Character* wielder, const Mobile::Character* target ) const
 {
   unsigned short dist = wielder->distance_to( target->toplevel_pos() );
-  INFO_PRINT_TRACE( 22 ) << "in_range(0x" << fmt::hexu( wielder->serial ) << ",0x"
-                         << fmt::hexu( target->serial ) << "):\n"
-                         << "dist:   " << dist << "\n"
-                         << "minrange: " << WEAPON_TMPL->minrange << "\n"
-                         << "maxrange: " << WEAPON_TMPL->maxrange << "\n"
-                         << "has_los:  " << wielder->realm()->has_los( *wielder, *target ) << "\n";
+  INFO_PRINTLN_TRACE( 22 )
```
frozenblit, line 462: looks funny. Does the double-call work?

frozenblit (approved):

pol-core/pol/item/weapon.cpp
turleypol: Yes tested and verified :)

PR #595, opened 2024-01-10
use new formatter for ERROR_PRINT, added ERROR_PRINTLN

turleypol:


pol-core/clib/cfgfile.cpp
```diff
@@ -986,27 +986,27 @@ void ConfigFile::display_error( const std::string& msg, bool show_curline,
...
     }
   }
 
   if ( show_curline )
-    tmp << "\tNear line: " << _cur_line << "\n";
+    tmp += fmt::format( "\n\tNear line: {}\n", _cur_line );
   if ( _element_line_start && !showed_elem_line )
-    tmp << "\tElement started on line: " << _element_line_start << "\n";
-  ERROR_PRINT << tmp.str();
+    tmp += fmt::format( "\n\tElement started on line: {}", _element_line_start );
```
KevinEady, line 1008: Could this cause two new-lines printed, if `show_curline` is true and the second conditional is false? There would be a trailing `\n` added to `tmp` and then an additional printed via `ERROR_PRINTLN`

pol-core/clib/streamsaver.cpp
```diff
@@ -53,7 +53,7 @@ OFStreamWriter::~OFStreamWriter()
         *_stream << _writer->c_str();
         _fs_time += t.ellapsed();
       }
-      ERROR_PRINT << "streamwriter " << _stream_name << " io time " << _fs_time.count( ) << "\n";
+      ERROR_PRINTLN("streamwriter {} io time {}",_stream_name,_fs_time.count( ));
```
KevinEady, line 56: hehe guess this didn't get formatted because of the `#if 0`, no big deal though

pol-core/pol/create.cpp
```diff
@@ -785,7 +771,7 @@ void ClientCreateCharKR( Network::Client* client, PKTIN_8D* msg )
       chr->equip( tmpitem );
     else
     {
-      ERROR_PRINT.Format( "Create Character: Failed to equip hair 0x{:X}\n" ) << tmpitem->graphic;
+      ERROR_PRINTLN( "Create Character: Failed to equip hair {:#X}", tmpitem->graphic );
```
KevinEady, line 774: Is there a difference between new fmt lib `{:#X}` and `{:#x}`? I've seen the lowercase version in the PR up until here

pol-core/pol/multi/multis.cpp
```diff
@@ -35,7 +35,7 @@ UMulti::UMulti( const Items::ItemDesc& itemdesc ) : Item( itemdesc, Core::UOBJ_C
 
   if ( !MultiDefByMultiIDExists( itemdesc.multiid ) )
   {
-    ERROR_PRINT.Format( "Tried to create a Multi type 0x{:X}\n" ) << itemdesc.objtype;
+    ERROR_PRINTLN( "Tried to create a Multi type {:#X}", itemdesc.objtype );
     throw std::runtime_error( "Unvalid Multi type" );
```
KevinEady, line 39: not part of PR but loving this "Unvalid" 😅

KevinEady (commented):
Some comments

pol-core/pol/create.cpp
turleypol: As mentioned in discord X is all uppercase x is all lowercase. I would suggest because it's already inconsistent that we decide which version we want and once every logger is adapted I perform a global replace.

KevinEady (approved):
:+1:

PR #598, opened 2024-01-14
POLLOG* macros use the new formatting

turleypol:
removed all old logging functions


pol-core/uotool/UoToolMain.cpp
```diff
@@ -112,31 +112,35 @@ unsigned char buffer[10000];
...
+                    static_cast<unsigned long>( tile.anim ) );
   if ( tile.height )
-    _tmp.Format( "  height: 0x{:02X}\n" ) << (int)tile.height;
+    fmt::format_to( std::back_inserter( _tmp ), "  height: {:#04X}\n", (int)tile.height );
   if ( tile.unk14 )
-    _tmp.Format( "  unk14:  0x{:02X}\n" ) << (int)tile.unk14;
+    fmt::format_to( std::back_inserter( _tmp ), "  unk14:  {:#04X}\n", (int)tile.unk14 );
   if ( tile.unk15 )
-    _tmp.Format( "  unk15:  0x{:02X}\n" ) << (int)tile.unk15;
+    fmt::format_to( std::back_inserter( _tmp ), "  unk15:  {:#04X}\n", (int)tile.unk15 );
```
KevinEady, line 143: The objtype format on 116 from `04X` to `06X` makes sense (since max objtype is >0xffff) but why are all these others changed from `02X` to `04X` ...?

KevinEady (approved):
👍

pol-core/bscript/compctx.cpp
```diff
@@ -146,16 +146,6 @@ int CompilerContext::skipcomments()
 
   return 0;
 }
-
-void CompilerContext::printOn( std::ostream& os ) const
```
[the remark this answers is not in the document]
turleypol: Removed it also in the header, thanks for spotting

pol-core/uotool/UoToolMain.cpp
turleypol: The padding with 0 includes the 0x so I had to add 2 for each format

PR #599, opened 2024-01-18
Prevent sending name on singleclick on hidden character

Kukkimonsuta:
When player1 sends 0x09 single click packet with serial of player2 who is nearby but hidden a name is sent back revealing information that there is player2 hidden nearby. This PR fixes the issue by using `is_visible_to_me` instead of `!is_concealed_from_me`. Range check is done withing `is_visible_to_me` so it was removed from the main condition as well.


turleypol:
Looking good, but please add core-changes

KevinEady (approved):

KevinEady (approved):

PR #614, opened 2024-01-30
CMake target cleanup, removed configure options

turleypol:
every 3rd party lib is now excluded from "all" build, meaning it gets 
only build when actually used.
removed the cmake option to partially build, due to the above changes
this is no more needed, just build the target
streamlined 3rdparty target names, they all start with lib now


KevinEady (approved):
LGTM! But are all the comments in `cmake/Boost.cmake` necessary? Can they be removed?

PR #623, opened 2024-02-08
Add proxy protocol support

Kukkimonsuta:
This change adds option to create listeners with proxy protocol v2 support (<https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt>). This allows for preserving client ips while hiding your actual game server behind proxies (ex. <https://www.haproxy.org/>).

To accept proxy protocol `Listener` in `uoclient.cfg` needs to be configured with `AllowProxy x.x.x.x`. This will cause listener to expect proxy protocol v2 from these IPs.

For example:
```
Listener
{
    port 5000
    Encryption ignition
    AllowProxy 10.0.1.14
}
```
- if incoming connection from 10.0.1.14 is using proxy protocol, it is accepted and for all further processing the IP address presented by the protocol is used instead
- if incoming connection from 10.0.1.14 is not using proxy protocol, it is rejected
- if incoming connection from 10.0.1.55 is using proxy protocol, it is rejected (since pol expect uo protocol)
- if incoming connection from 10.0.1.55 is not using proxy protocol, it is accepted

This change also adds `ProxyMatch` property to `GameServer` in `servers.cfg`. When using proxy protocol this property should be present on at least one game server, otherwise clients connecting through proxy will receive empty server list. This may seem a little unintitive, however it provides a bit more options to directing users to correct listeners (can have open "secret" public ip as well as proxied one while not automatically exposing the former to proxy users).
```
GameServer
{
    Name        My server
    IP          89.100.100.65
    Port        5000
}

GameServer
{
    Name        My proxied server
    IP          89.100.100.65
    Port        5000
    ProxyMatch  10.0.1.14
}
```
- client connecting through proxy `10.0.1.14` will see only `My proxied server`
- client connecting through proxy `10.0.1.22` won't see any servers
- client connecting without proxy will see only `My server`

Proxy information is not exposed to scripting.

When client (172.17.0.1) is connecting through a proxy (127.0.0.1) log looks like this:
```
[02/14 16:36:05] Client#1 connected from 127.0.0.1 (1/0 connections) on interface 127.0.0.1
[02/14 16:36:05] Client#1 (172.17.0.1): connected through proxy 127.0.0.1
[02/14 16:36:05] Account admin logged in from 172.17.0.1
[02/14 16:36:09] Client#1 (172.17.0.1): disconnected (account admin)
[02/14 16:36:10] Client#2 connected from 127.0.0.1 (1/0 connections) on interface 127.0.0.1
[02/14 16:36:10] Client#2 (172.17.0.1): connected through proxy 127.0.0.1
[02/14 16:36:10] Account admin logged in from 172.17.0.1
[02/14 16:36:15] Account admin selecting character kukkino
[02/14 16:36:22] Client#2 (172.17.0.1): disconnected (account admin)
```

When client (127.0.0.1) is connecting without proxy when proxy protocol is expected, log looks like this:
```
[02/14 16:41:11] Client#4 connected from 127.0.0.1 (1/0 connections) on interface 127.0.0.1
[02/14 16:41:11] Client#4 (127.0.0.1): disconnected due to invalid proxy header
[02/14 16:41:11] Client#4 (127.0.0.1): disconnected (account unknown)
```

### DIFF
--- a/cmake/compile_defs.cmake
+++ b/cmake/compile_defs.cmake
@@ -42,7 +42,8 @@ function(set_compile_flags target is_executable)
       WIN32
       NOMINMAX
       WINDOWS
-      WINVER=0x0501
+      WINVER=0x0601
+      _WIN32_WINNT=0x0601
       _CONSOLE
     )
     if (${ARCH_BITS} EQUAL "64")
--- a/docs/docs.polserver.com/pol100/configfiles.xml
+++ b/docs/docs.polserver.com/pol100/configfiles.xml
@@ -109,7 +109,7 @@ Listener
     KeepClients      (integer 0/1 {default 0})
     Encryption       (string encryption type {default none})
     AOSResistances   (integer 0/1)
-    UseProxyProtocol (integer 0/1 {default 0})
+    [AllowProxy      (IPaddrr)[/(netmask)]]...
 }
 [Listener ...]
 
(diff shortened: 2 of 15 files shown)


Kukkimonsuta:
Updated to use `AllowProxy x.x.x.x` instead of enabling proxy protocol through entire listener.

frozenblit (approved):

frozenblit (approved):

PR #630, opened 2024-03-06
Add daemon watch mode to ecompile

KevinEady:
Using the `-W` switch with ecompile will now enter a "watch mode", re-compiling sources on file modification (including sources changed via a modification to a dependent include or module).

TODO:
- [x] ~~add ecompile.cfg option to use OS-independent generic watcher backend for efsw library?~~ not necessary
- [x] docs
- [x] tests
- [x] core changes


pol-core/ecompile/ECompileMain.cpp
```diff
@@ -138,6 +145,14 @@ struct Comparison
 Compiler::SourceFileCache em_parse_tree_cache( summary.profile );
 Compiler::SourceFileCache inc_parse_tree_cache( summary.profile );
 
+using DependencyInfo = std::map<fs::path, std::set<fs::path>>;
+// Map owner (sources) -> dependencies
+DependencyInfo owner_dependencies;
+// Map dependency -> owners (sources)
+DependencyInfo dependency_owners;
+
```
turleypol, line 153: You need these globals only because you added code into process_file_wrapper. If you put this code into the calling function these variables can be local.

pol-core/ecompile/ECompileMain.cpp
```diff
@@ -417,6 +506,16 @@ void process_file_wrapper( const char* path )
     if ( !keep_building )
       throw;
   }
+
+  if ( watch_source )
```
turleypol, line 510: See comment at globals

pol-core/ecompile/ECompileMain.cpp
```diff
@@ -713,6 +820,250 @@ void parallel_process( const std::set<std::string>& files )
...
+    {
+      auto dirpath = fs::canonical( dir );
+      if ( listener.add_watch_dir( dirpath ) )
+      {
+        VERBOSE_PRINTLN( "Add watch dir {}", dirpath );
+      }
+    }
+  };
+
+  for ( const auto& dep_owners : dependency_owners )
```
turleypol, line 956: I think always empty at start

pol-core/ecompile/ECompileMain.cpp
```diff
@@ -713,6 +820,250 @@ void parallel_process( const std::set<std::string>& files )
...
+  {
+    listener.add_file( dep_owners.first );
+  }
+
+  for ( const auto& directory : compiled_dirs )
```
turleypol, line 961: Same here

pol-core/ecompile/ECompileMain.cpp
```diff
@@ -713,6 +820,250 @@ void parallel_process( const std::set<std::string>& files )
...
+        Tools::Timer<> timer;
+        for ( const auto& filepath : to_compile )
+        {
+          std::set<fs::path> removed_dependencies;
+          std::set<fs::path> new_dependencies;
+          try
+          {
+            process_file_wrapper( filepath.generic_string().c_str(), &removed_dependencies,
+                                  &new_dependencies );
+          }
```
turleypol, line 1027: Call here add_dependencies_info

pol-core/ecompile/EfswFileWatchListener.cpp
```diff
@@ -0,0 +1,175 @@
...
+    break;
+
+  default:
+    break;
+  }
+}
+
+// Adds a file to watched files. Returns `true` if the file was newly added to
+// `watched_files`.
+bool EfswFileWatchListener::add_file( const fs::path& filepath )
```
turleypol, line 72: const ref plus .emplace still gives you a copy. Have plain fs::path + std::move for emplace removes the extra copy

pol-core/ecompile/EfswFileWatchListener.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+  }
+}
+
+// Removes the messages and inserts them into the provided `to_message` list.
+// The messages are only removed if the `handle_messages_by` time clock has
+// passed.
+void EfswFileWatchListener::take_messages( std::list<WatchFileMessage>& to_messages )
+{
+  std::lock_guard<std::mutex> lock( mutex );
```
turleypol, line 168: conditional_variable would remove the current "busy" loop with the sleep of 1s

pol-core/ecompile/ECompileMain.cpp
KevinEady: These vars are also used in `add_dependency_info`  as well as `EnterWatchMode`.

pol-core/ecompile/ECompileMain.cpp
KevinEady: `dependency_owners` gets filled during compilation, and then get watched in `EnterWatchMode`.

pol-core/ecompile/ECompileMain.cpp
KevinEady: `compiled_dirs` gets filled during compilation inside `recurse_collect`, and then get watched in `EnterWatchMode`.

pol-core/ecompile/ECompileMain.cpp
KevinEady: `add_dependency_info` is inside `process_file_wrapper `, so that when a file is compiled, dependency info is updated. Are you suggesting to move it?

pol-core/ecompile/EfswFileWatchListener.cpp
KevinEady: We still need the listener for Ctrl+C which is why there's the 1s wait. Can that be achieved with cond vars?

pol-core/ecompile/ECompileMain.cpp
KevinEady: `watch_source` is read in two places: (1) `process_file_wrapper` to determine if `add_dependency_info` should be called, (2) end of `run()` to determine if `EnterWatchMode` should be called. I think`watch_source` should remain global.

pol-core/ecompile/EfswFileWatchListener.cpp
KevinEady: Found `cv.wait_for`, :+1:

KevinEady: Resolved in https://github.com/polserver/polserver/pull/630/commits/c5a2fa7f7b553a333c43fb5f1d48526bb2f88807

turleypol (approved):

PR #633, opened 2024-03-15
Add `item.weight_multiplier_mod`

KevinEady:
TODO:
- [x] implementation
- [x] tests
- [x] docs
- [x] core-changes

### DIFF
--- a/pol-core/pol/uoscrobj.cpp
+++ b/pol-core/pol/uoscrobj.cpp
@@ -1103,6 +1103,9 @@ BObjectImp* Item::get_script_member_id( const int id ) const
     return new BError( "This item is not owned by any character" );
     break;
   }
+  case MBR_WEIGHT_MULTIPLIER_MOD:
+    return new Double( weight_multiplier_mod() );
+    break;
   default:
     return nullptr;
   }
@@ -1572,6 +1575,8 @@ BObjectImp* Item::set_script_member_id( const int id, int value )
   case MBR_NO_DROP:
     no_drop( value ? true : false );
     return new BLong( no_drop() );
+  case MBR_WEIGHT_MULTIPLIER_MOD:
+    return set_script_member_id_double( id, value );
   default:
     return nullptr;
   }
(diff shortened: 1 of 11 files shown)


pol-core/pol/uoscrobj.cpp
```diff
@@ -1595,6 +1600,37 @@ BObjectImp* Item::set_script_member_id_double( const int id, double value )
...
+    {
+      return new BError( "weight_multiplier_mod cannot be less than 0.0" );
+    }
+
+    if ( container )
+    {
+      int current_weight = weight();
+      int will_be_new_weight = weight_of( amount_, value );
+      int weight_delta = will_be_new_weight - current_weight;
+      container->add_bulk( 0, weight_delta );
```
turleypol, line 1615: weight_delta can be negative, but add_bulk executes a `+= static_cast<u16>()` with it (wtf)
add_bulk needs to do something like:
`int newweight = held_weight_; newweight+=weight_delta; held_weight_=clamp_convert<u16>(newweight);`

pol-core/pol/uoscrobj.cpp
```diff
@@ -1595,6 +1600,37 @@ BObjectImp* Item::set_script_member_id_double( const int id, double value )
...
+  {
+    if ( value < 0.0 )
+    {
+      return new BError( "weight_multiplier_mod cannot be less than 0.0" );
+    }
+
+    if ( container )
+    {
+      int current_weight = weight();
+      int will_be_new_weight = weight_of( amount_, value );
```
turleypol, line 1613: That seems to be the only spot where the new weight_of method is called. Since you store the original weight before why not directly set the new weight? This would remove the need of the new function.

pol-core/pol/uoscrobj.cpp
KevinEady: Addressed in [288f7f4](https://github.com/polserver/polserver/pull/633/commits/288f7f4e923618eb3551843b0bb804644c5c154a).

pol-core/pol/uoscrobj.cpp
KevinEady: Addressed in [288f7f4](https://github.com/polserver/polserver/pull/633/commits/288f7f4e923618eb3551843b0bb804644c5c154a).

turleypol (approved):

PR #640, opened 2024-04-15
Add servspecopt BoatSailsCollide

KevinEady:


pol-core/pol/multi/boat.cpp
```diff
@@ -669,6 +681,9 @@ void UBoat::regself()
   const MultiDef& md = multidef();
   for ( const auto& ele : md.hull )
   {
+    if ( objtype_passable( ele->objtype ) )
```
turleypol, line 684: Since this skips the global_hull insert the method unregself should also skip the remove. Otherwise if could delete the hull entry of a different boat.

KevinEady: Addressed in [68c40c0](https://github.com/polserver/polserver/pull/640/commits/68c40c06f9327369ca74ea83a9eba74707e3f4ce)

turleypol (approved):

PR #656, opened 2024-07-10
Boat refactor Part 3 of N

turleypol:
refactored boat move, fixed truncation of position when switching realm.
more simplifications
added more tests

Getting to an end I guess.. :)

### DIFF
--- a/pol-core/pol/uworld.cpp
+++ b/pol-core/pol/uworld.cpp
@@ -84,10 +84,10 @@ void remove_multi_from_world( Multi::UMulti* multi )
   zone.multis.erase( itr );
 }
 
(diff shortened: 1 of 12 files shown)


pol-core/pol/uworld.cpp
```diff
@@ -99,10 +98,10 @@ void move_multi_in_world( unsigned short oldx, unsigned short oldy, unsigned sho
     newzone.multis.push_back( multi );
   }
 
-  if ( multi->realm() != oldrealm )
+  if ( multi->realm() != newpos.realm() )
```
frozenblit, line 101: I wonder if there's any side-effect of changing this from oldpos to newpos. Technically before the multi->realm() was already changed before this function was called, now it needs to be called afterwards

frozenblit (approved):

PR #660, opened 2024-07-13
rewrite Decay system

turleypol:
removed setting to switch to multithread decay
Syshook:: call_long treats true/false as 1/0
can_decay hook if it returns 2 skip multicheck

- [x] docs

### DIFF
--- a/pol-core/pol/syshookscript.cpp
+++ b/pol-core/pol/syshookscript.cpp
@@ -104,7 +104,7 @@ bool ExportScript::call( unsigned PC, BObjectImp* p0, BObjectImp* p1, BObjectImp
 
     uoexec.exec();
 
-    bool istrue = check_bool();
+    bool istrue = expect_bool();
 
     // delete current state and reenable backup
     LoadStack( backup );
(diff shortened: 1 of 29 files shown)


pol-core/pol/syshookscript.cpp
```diff
@@ -604,5 +466,58 @@ size_t ExportScript::estimateSize() const
...
+std::string ExportScript::check_string()
+{
+  if ( uoexec.error() || uoexec.ValueStack.empty() )
+    return "error";
+  auto ret = uoexec.ValueStack.back().get()->impptr()->getStringRep();
+  uoexec.ValueStack.pop_back();
+  return ret;
+}
+
+BObjectImp* ExportScript::check_imp()
```
frozenblit, line 511: I wonder if this might leak at the call sites, in case something goes wrong between calling the method and returning the BObject for it

pol-core/pol/syshookscript.cpp
```diff
@@ -604,5 +466,58 @@ size_t ExportScript::estimateSize() const
 {
   return sd.estimatedSize() + uoexec.sizeEstimate();
 }
+
+bool ExportScript::check_bool()
```
frozenblit, line 470: I'm not sure if "check" is the best name for the methods, because it implies you're only validating the result but you actually read it. Did you consider other names?

Maybe an "expect_bool", "expect_int", etc could imply you read the result from the script with some expectation about the type

frozenblit (approved):

pol-core/pol/syshookscript.cpp
turleypol: There is hope :)

PR #662, opened 2024-07-15
Add stacktrace to Error instance

KevinEady:
TODO:
- [x] handle async situations, eg. `OpenConnection`
- [ ] async tests
- [x] fix tests for windows
- [ ] docs


pol-core/bscript/berror.cpp
```diff
@@ -127,5 +133,123 @@ BObjectImp* BError::array_assign( BObjectImp* /*idx*/, BObjectImp* /*target*/, b
...
+  }
+
+  EScriptProgram* prog = const_cast<EScriptProgram*>( exec->prog() );
+  script_.set( prog );
+
+  if ( !stack_.empty() )
+  {
+    return;
+  }
+  std::vector<ReturnContext> callStack = exec->ControlStack;
```
turleypol, line 231: Since BError is one of the most created types I fear a bit the performance impact. Maybe have a setting for it?
In general this method could have some performance improvements. Don't copy ControlStack and instead reserve the size+1 add current PC and then iterate backwards over the ControlStack and add the pc.

pol-core/bscript/berror.cpp
```diff
@@ -127,5 +133,123 @@ BObjectImp* BError::array_assign( BObjectImp* /*idx*/, BObjectImp* /*target*/, b
...
+    ReturnContext& rc = callStack.back();
+    stack_.push_back( rc.PC );
+    callStack.pop_back();
+  }
 }
+
+size_t BError::sizeEstimate() const
+{
+  size_t size = base::sizeEstimate() + sizeof( ref_ptr<EScriptProgram> );
+  size += 3 * sizeof( unsigned int* ) + stack_.capacity() * sizeof( unsigned int );
```
turleypol, line 251: Luckily no longer needed to write it this way

KevinEady:
Closed by #739 for a middle-ground design, giving no additional overhead if feature is not used.

PR #664, opened 2024-07-16
Add source location to `Print()` messages sent to debugger

KevinEady:
The log message includes the source location of the `Print()` call.

### DIFF
--- a/pol-core/pol/dap/clientthread.cpp
+++ b/pol-core/pol/dap/clientthread.cpp
@@ -69,6 +69,21 @@ void DebugClientThread::on_print( const std::string& output )
   dap::OutputEvent event;
   event.category = "stdout";
   event.output = output + "\n";
+
+  // The `ExecutorDebugListener` methods are called from the script thread
+  // (which runs inside a PolLock), so no need to lock. Additionally, the
+  // `UOExecutor` should always exist, as these methods are all called within an
+  // `Executor` instance method, but check just in case. Not a real performance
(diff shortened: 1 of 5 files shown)


pol-core/pol/dap/clientthread.cpp
```diff
@@ -69,6 +69,21 @@ void DebugClientThread::on_print( const std::string& output )
...
+  // The `ExecutorDebugListener` methods are called from the script thread
+  // (which runs inside a PolLock), so no need to lock. Additionally, the
+  // `UOExecutor` should always exist, as these methods are all called within an
+  // `Executor` instance method, but check just in case. Not a real performance
+  // hit, since this is only called when printing with the debugger is attached.
+  if ( _uoexec_wptr.exists() )
+  {
+    UOExecutor* exec = _uoexec_wptr.get_weakptr();
+    dap::Source source;
+    source.path = _script->dbg_filenames[_script->dbg_filenum[exec->PC]];
```
frozenblit, line 82: do the array elements always exist?

KevinEady: They should ... We only successfully attach the debugger if the debug info can be loaded. The elements would only not exist if the .ecl and .dbg were out of sync.

FWIW, we do the same thing in other places, eg.:
https://github.com/polserver/polserver/blob/271722af13611ac0f8539ed18828eb407f765782/pol-core/pol/dap/clientthread.cpp#L565

frozenblit (approved):

PR #667, opened 2024-07-20
Add filter method to escript array object

KevinEady:
- Add `BContinuation` structure to allow core to call user functions asynchronously (ie. _not_ as an exported script hook).
- Add filter method to escript array object using continuation approach.


pol-core/bscript/object.cpp
```diff
@@ -1653,6 +1655,85 @@ BObjectImp* ObjArray::call_method_id( const int id, Executor& ex, bool /*forcebu
...
+      - Call the user function again by returning a continuation
+      - Return something else to provide that value back to the script.
+      */
+      auto callback = [this, filteredObj = BObjectRef( new ObjArray ),
+                       processedObj = BObjectRef( new BLong( 1 ) )](
+                          Executor& ex, BContinuation* continuation,
+                          BObjectRef result ) -> BObjectImp*
+      {
+        auto filtered = static_cast<ObjArray*>( filteredObj->impptr() );
+        auto processed = static_cast<BLong*>( processedObj->impptr() );
```
turleypol, line 1704: I think someday we should add to BObject
template <typename T>
T* ptr() {return static_cast<T*>(objimp.get());}
To not being forced to write everywhere such long lines.

pol-core/bscript/object.cpp
```diff
@@ -1653,6 +1655,85 @@ BObjectImp* ObjArray::call_method_id( const int id, Executor& ex, bool /*forcebu
...
+
+        // Do something with result.
+        // If the result is true, add it to the filtered array.
+        if ( result->isTrue() )
+        {
+          filtered->ref_arr.push_back( ref_arr[processed->value() - 1] );
```
turleypol, line 1710: We hope that the filter function does not modify the array itself. Makes not really sense to do it, performing a copy of the ref_arr would solve it, but could be costly depending how big the data is.
Maybe simply checking if processed->value()-1 is still in the bounds of the array is enough

pol-core/bscript/object.cpp
```diff
@@ -1984,7 +2065,7 @@ BObjectImp* BFunctionRef::call_method( const char* methodname, Executor& ex )
 
 bool BFunctionRef::validCall( const int id, Executor& ex, Instruction* inst ) const
 {
-  if ( id != MTH_CALL )
+  if ( id != -1 && id != MTH_CALL )
```
turleypol, line 2068: Why not use for continuation also the MTH_CALL id. In the end you want to call it, so why add a special handling for it

pol-core/bscript/executor.cpp
```diff
@@ -3685,5 +3761,29 @@ unsigned long Executor::GetTimeUs()
...
+    args.push_back( BObjectRef( new BObject( UninitObject::create() ) ) );
+  }
+
+  // Resize and erase the state in `args` since it was moved above.
+  args.resize( func->numParams() );
+
+  // Move all arguments to the value stack
+  fparams.insert( fparams.end(), std::make_move_iterator( args.begin() ),
+                  std::make_move_iterator( args.end() ) );
+  args.erase( args.begin(), args.end() );
```
turleypol, line 3784: .clear();

pol-core/bscript/executor.inl.h
```diff
@@ -0,0 +1,67 @@
...
+  {
+    args.push_back( BObjectRef( new BObject( UninitObject::create() ) ) );
+  }
+  // Resize and erase the state in `args` since it was moved above.
+  args.resize( func->numParams() );
+
+  // Move all arguments to the value stack
+  ValueStack.insert( ValueStack.end(), std::make_move_iterator( args.begin() ),
+                     std::make_move_iterator( args.end() ) );
+  args.erase( args.begin(), args.end() );
```
turleypol, line 58: Same problem like with continuation.
Instead of defaulting to an empty array: use std::optional or pass a pointer which can be null

pol-core/bscript/object.cpp
```diff
@@ -1653,6 +1655,99 @@ BObjectImp* ObjArray::call_method_id( const int id, Executor& ex, bool /*forcebu
...
+      // We pass to the lambda a reference to the element in case the user
+      // function modifies ref_arr.
+      //
+      // Returns a `BObjectImp`:
+      // - Call the user function again by returning the same continuation via
+      //   `ex.withContinuation`.
+      // - Return something else (in this case, the filtered array) to provide
+      //   that value back to the script.
+      auto callback = [this, filteredRef = BObjectRef( new ObjArray ),
+                       processedRef = BObjectRef( new BLong( 1 ) ),
```
turleypol, line 1703: Since you create a new objectref for each run, so that the script cannot modify the index, wouldn't it be enough to just store/increment a plain integer?

KevinEady: yes, but i couldn't figure out how to attach a "state member" to the lambda, eg:

```cpp
int processed = 0;

auto callback = [this, processed = std::move(processed)] () {
    processed++;
}
```
but this does not work

KevinEady: > Since you create a new objectref for each run,

Also, is it really creating a new objectref for each run? I thought with eg. `processedRef = BObjectRef(...)` would only create this object when the lambda itself is created, and is "attached" to the lambda (and when the lambda gets freed, as do the bobjectrefs)

pol-core/bscript/bobject.h
```diff
@@ -116,6 +116,7 @@ class BObjectImp : public ref_counted
     OTFuncRef = 39,
     OTExportScript = 40,
     OTStorageArea = 41,
+    OTContinuation = 42,
```
turleypol, line 119: Since it's only internally used and has no script representation would be better to leave don't know 100 gap and add a comment that's it's internal. Otherwise we will wonder after a few weeks and when we add a new script object the scripters will wonder where TypeOfInt 42 is

pol-core/bscript/executor.cpp
```diff
@@ -2667,7 +2729,12 @@ void Executor::ins_call_method( const Instruction& ins )
   std::string name( strm.str() );
   unsigned long profile_start = GetTimeUs();
 #endif
-  BObjectImp* imp = ValueStack.back()->impptr()->call_method( ins.token.tokval(), *this );
+  BObjectImp* imp;
+
+  if ( strcmp( ins.token.tokval(), "impptr" ) == 0 )
```
KevinEady, line 2734: remove leftover debugging

turleypol (approved):

PR #668, opened 2024-07-20
Moved target/menu members from char to ClientGameData, simplified memory size estimation

turleypol:
These members only work with an active client.
Moved in few classes members to get rid of padding.
Main part is usage of new helper functions to estimate the memory size of std containers.

### DIFF
--- a/pol-core/pol/item/itemdesc.cpp
+++ b/pol-core/pol/item/itemdesc.cpp
@@ -911,8 +911,7 @@ size_t ItemDesc::estimatedSize() const
   size_t size = sizeof( ItemDesc ) + objtypename.capacity() + tooltip.capacity() +
                 walk_on_script.estimatedSize() + on_use_script.estimatedSize() +
                 control_script.estimatedSize() + create_script.estimatedSize() +
-                destroy_script.estimatedSize() + 3 * sizeof( ResourceComponent* ) +
-                resources.capacity() * sizeof( ResourceComponent ) + props.estimatedSize();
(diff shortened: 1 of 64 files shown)


pol-core/pol/item/itemdesc.cpp
```diff
@@ -911,8 +911,7 @@ size_t ItemDesc::estimatedSize() const
   size_t size = sizeof( ItemDesc ) + objtypename.capacity() + tooltip.capacity() +
                 walk_on_script.estimatedSize() + on_use_script.estimatedSize() +
                 control_script.estimatedSize() + create_script.estimatedSize() +
-                destroy_script.estimatedSize() + 3 * sizeof( ResourceComponent* ) +
-                resources.capacity() * sizeof( ResourceComponent ) + props.estimatedSize();
+                destroy_script.estimatedSize() + Clib::memsize( resources ) + props.estimatedSize();
   size += 3 * sizeof( void* );
   for ( const auto& ignore : ignore_cprops )
   {
```
KevinEady, line 917: If I understand the PR correctly, could this section be changed to `Clib::memsize( ignore_cprops )` since `ignore_cprops` is a `std::set<std::string>` and would be covered by the template?

KevinEady (approved):
LGTM!

I did a quick search in codebase for `3 * sizeof` and there were a few remaining hits, perhaps they were overlooked or just better to leave as current implementation? (see example in one comment)

pol-core/pol/item/itemdesc.cpp
turleypol: As discussed: followup PR

PR #669, opened 2024-07-20
Fix status hp values upon realm change

Kukkimonsuta:
When changing realms player gets sent status update packet with values scaled to 1000 instead of actual values. This PR changes behavior so that full status update is sent, which is consistent with all other usages - `Core::send_short_statmsg` is otherwise used only when sending status of other entities that aren't current player.

### DIFF
--- a/docs/docs.polserver.com/pol100/corechanges.xml
+++ b/docs/docs.polserver.com/pol100/corechanges.xml
@@ -2,9 +2,14 @@
 <ESCRIPT>
 	<header>
 		<topic>Latest Core Changes</topic>
-		<datemodified>07-16-2024</datemodified>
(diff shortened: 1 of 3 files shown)


frozenblit:
Can you add the core-changes entry? Otherwise looks good to me. 

I wonder why it wasn't caught before - did anything change recently?

Kukkimonsuta:
Core changes updated.

I think it's because you don't really change realms that often and it fixes itself as soon as you receive another update from other source. Also it's somewhat conditional as well - for example if player using ClassicUO has enabled hp "line" under characters he won't even notice as the client requests another update immediately.

frozenblit (approved):

PR #671, opened 2024-07-21
Add compiler support for function expressions

KevinEady:
Add compiler support for function expressions via syntax `@(param1, param2, ..., paramN) { ... }`.


pol-core/bscript/compiler/ast/FunctionExpression.cpp
```diff
@@ -0,0 +1,26 @@
...
+}
+
+void FunctionExpression::accept( NodeVisitor& visitor )
+{
+  visitor.visit_function_expression( *this );
+}
+
+void FunctionExpression::describe_to( std::string& w ) const
+{
+  fmt::format_to( std::back_inserter( w ), "boolean-value({})", true );
```
turleypol, line 23: That looks like c&p error :)

KevinEady: Yeah it's a placeholder until I figure out how to properly describe it 👍

pol-core/bscript/executor.cpp
```diff
@@ -3110,7 +3115,30 @@ void Executor::ins_bitwise_not( const Instruction& /*ins*/ )
 void Executor::ins_funcref( const Instruction& ins )
 {
   ValueStack.push_back( BObjectRef(
-      new BObject( new BFunctionRef( ins.token.lval, ins.token.type, scriptname() ) ) ) );
+      new BObject( new BFunctionRef( ins.token.lval, ins.token.type, scriptname(), {} ) ) ) );
+}
+
+void Executor::ins_functor( const Instruction& ins )
+{
+  int parameter_count = static_cast<BLong*>( ValueStack.back()->impptr() )->value();
```
turleypol, line 3123: passert_always(ValueStack.back()->Isa(Blong));
Would I think not hurt and shows that something is broken instead of complete undefined behaviour. Two lines later I suggest the same

turleypol (dismissed):

turleypol (approved):

PR #673, opened 2024-07-28
improved escript formatter when formatting if's/function parameters

turleypol:
logical operators get now a preferred break style same for commas in function/program parameters. split based on this flag the line


KevinEady (approved):
LGTM! See Discord for some discussions

KevinEady (approved):

PR #686, opened 2024-08-05
Add function reference table to EScriptProgram

KevinEady:
Encapsulates the function reference attributes (parameter count, capture count, is variadic) into an `EPFunctionReference` structure used by the executor.

### DIFF
--- a/pol-core/bscript/compiler/codegen/FunctionReferenceRegistrar.h
+++ b/pol-core/bscript/compiler/codegen/FunctionReferenceRegistrar.h
@@ -1,0 +1,40 @@
+#pragma once
+
+#include <map>
+#include <memory>
+#include <string>
+#include <vector>
+
+namespace Pol::Bscript::Compiler
+{
+class FlowControlLabel;
+class UserFunction;
+class FunctionReferenceDescriptor;
+
+class FunctionReferenceRegistrar
+{
+public:
+  FunctionReferenceRegistrar();
+
+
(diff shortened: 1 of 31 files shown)


pol-core/bscript/compiler/codegen/FunctionReferenceRegistrar.h
```diff
@@ -0,0 +1,40 @@
...
+
+class FunctionReferenceRegistrar
+{
+public:
+  FunctionReferenceRegistrar();
+
+
+  std::vector<FunctionReferenceDescriptor> take_descriptors();
+
+  void lookup_or_register_reference( const UserFunction& node, unsigned& reference_index );
```
turleypol, line 22: Returning the reference_index would make the calling side clearer

turleypol (commented):
Just one small comment you could consider. If corechanges are added I would approve it. (Due to version change we need one)

pol-core/bscript/compiler/codegen/FunctionReferenceRegistrar.h
KevinEady: I wanted to keep with the existing prototype style: https://github.com/polserver/polserver/blob/e8b5c15ca7bdb3dff22ca66795e71413b21b6adf/pol-core/bscript/compiler/codegen/ModuleDeclarationRegistrar.h#L25-L26

turleypol (approved):

PR #692, opened 2024-08-12
Delayed AST building of class variable statements

KevinEady:
### Split variable statements into two passes

1. When the `SourceFileProcessor` (first-pass AST visitor) visits a class declaration, it will add an (empty) `ClassBody` node to top-level statements. This node is stored in the `AvailableParseTree` (which the `FunctionResolver` keeps as a list of ASTs to visit in second-pass)
2. When the `UserFunctionVisitor` (second-pass AST visitor) visits a class declaration, it adds the var-statement children to the (previously-empty) `ClassBody` node corresponding to its declaration.

Since the var statements are still in proper tree traversal order inside top-level statements (as children of `ClassBody` vs direct children of `TopLevelStatements`), all node visitors (semantic analysis, optimizer, instruction generation, ...) will continue to work. Additionally, any node visitor could decide _not_ to visit class var statements, but currently not the case.

----

### Add class references for scoped identifiers

A class that only provides variables has no functions, so we need to track their availability. The `FunctionResolver` now has new members `resolved_classes` and `unresolved_classes` for tracking classes. When visiting a scoped identifier, we force a reference to the class via `force_reference( ScopeName )`, and resolving the reference via `register_class_declaration( ClassDeclaration* )`

These are the "class equivalent" for the function-based `resolved_functions`, `unresolved_function_links`, `force_reference`, and `register_user_function` respectively.

### DIFF
--- a/pol-core/bscript/compiler/analyzer/SemanticAnalyzer.cpp
+++ b/pol-core/bscript/compiler/analyzer/SemanticAnalyzer.cpp
@@ -897,7 +897,8 @@ bool SemanticAnalyzer::report_function_name_conflict( const SourceLocation& refe
                                                       const std::string& function_name,
                                                       const std::string& element_description )
 {
-  return report_function_name_conflict( workspace, report, referencing_loc, function_name,
+  return report_function_name_conflict( workspace, report, referencing_loc,
+                                        ScopableName( current_scope_name, function_name ).string(),
                                         element_description );
 }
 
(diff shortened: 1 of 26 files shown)


KevinEady:
Timing information of compiling ModernDistro's **1468 scripts** with master's 36.49s vs this branch'ss 37.58s, an increase of  **1.09s** total / **0.742ms** per file / **2.98712%**

turleypol (approved):

PR #694, opened 2024-08-14
Add super() function call for constructors

KevinEady:
### Introduce `super()` function call
- Introduce `SuperFunction`, a child `UserFunction`. Right now there is no real reason for it to be a child .. it could just be a UserFunction but who knows maybe we'll need a dynamic cast later.
- Add UserFunctionType `Super`
- Create `ClassLink`, an analogy of `FunctionLink`
- A ClassDeclaration has links to it base class ClassDeclarations via ClassLink
- Introduce `GeneratedFunctionBuilder`, which builds a compiler-generated function (currently, only super)
- Add ability for FunctionResolver to resolve generated functions by passing nullptr as the parse-tree context.
- Introduce scopable names to function parameters (declaration side) and arguments (calling side), so we can resolve ctor parameters correctly

### Additional grammar and semantic changes
- Make class parameters required (as discussed)
- Make `return` statement in ctors automatically return `this`; error if return has a value inside a ctor

### Additional function resolver fixes
- Fix issue with explicit global function references `@::foo` being registered as an "unspecified" function reference (ie. `@foo`), which uses calling scope as context when resolving functions


pol-core/bscript/compiler/analyzer/SemanticAnalyzer.cpp
```diff
@@ -898,7 +1058,9 @@ void SemanticAnalyzer::visit_user_function( UserFunction& node )
   LocalVariableScope scope( local_scopes, node.local_variable_scope_info );
   visit_children( node );
   // We do not allow nested scope names
-  current_scope_name = ScopeName::Global;
+  // TODO fix this because of funcexprs
```
KevinEady, line 1061: can remove todo, was a self-reminder during dev

pol-core/bscript/compiler/ast/ClassDeclaration.h
```diff
@@ -30,8 +31,15 @@ class ClassDeclaration : public Node
...
   Node* class_body;
 
-  // Will be null if class has no constructor declared.
-  const std::shared_ptr<FunctionLink> constructor;
+  // Will have no function linked if class has no constructor defined.
+  const std::shared_ptr<FunctionLink> constructor_link;
+
+  // Filled by SemanticAnalyzer with functions owned in
+  // CompilerWorkspace::user_functions.
+  std::vector<std::shared_ptr<ClassLink>> base_class_links;
```
KevinEady, line 39: comment is wrong, done by generated function builder (was done in sem analyzer prior to refactor commit)

turleypol (approved):

PR #695, opened 2024-08-18
Support linking functions in super scope; explicit `super::` scope calls

KevinEady:
### Function resolver changes
- In the areas where we were checking "if calling_scope is not empty": instead of checking _only_ the calling scope, check the calling scope + ancestors
- In the case of a `super::` call, we do _not_ check current calling scope, but only ancestors.

### DIFF
--- a/pol-core/bscript/compiler/astbuilder/FunctionResolver.cpp
+++ b/pol-core/bscript/compiler/astbuilder/FunctionResolver.cpp
@@ -226,12 +226,13 @@ bool FunctionResolver::resolve( std::vector<AvailableParseTree>& to_build_ast )
         if ( !calling_scope.empty() )
         {
           std::set<std::string> visited;
-          std::list<std::string> to_link( { calling_scope } );
+          std::list<std::string> to_check( { calling_scope } );
           bool handled = false;
 
-          for ( auto itr = to_link.begin(); !handled && itr != to_link.end(); itr++ )
+          for ( auto to_check_itr = to_check.begin(); !handled && to_check_itr != to_check.end();
+                ++to_check_itr )
           {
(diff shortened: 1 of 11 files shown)


pol-core/bscript/compiler/astbuilder/FunctionResolver.cpp
```diff
@@ -549,15 +550,15 @@ bool FunctionResolver::build_if_available( std::vector<AvailableParseTree>& to_b
 
       if ( handled_by_scope( cd->name ) )
       {
-        itr = to_link.end();
+        to_check_itr = to_check.end();
```
turleypol, line 553: Not needed

turleypol (approved):

pol-core/bscript/compiler/astbuilder/FunctionResolver.cpp
KevinEady: Ah yup this was leftover from previous implementation

pol-core/bscript/compiler/astbuilder/FunctionResolver.cpp
```diff
@@ -471,16 +512,55 @@ bool FunctionResolver::build_if_available( std::vector<AvailableParseTree>& to_b
...
+
+      auto cd = cd_itr->second;
+      if ( visited.find( cd->name ) != visited.end() )
+        continue;
+
+      visited.insert( cd->name );
+
+      if ( handled_by_scope( cd->name ) )
+      {
+        to_check_itr = to_check.end();
```
KevinEady, line 553:
```suggestion
```

turleypol (approved):

PR #696, opened 2024-08-18
more formatting improvements

turleypol:
replace inside of comments tabs with spaces if UseTabs is false
do not strip whitespaces at the beginning of linecomments
better handling of long lines eg string additions
"binpack":
  if an array fits equally into up to 3 lines use this instead of having each element in a newline
  struct/dicts get only packed if they fit into a single line, or half of the room for two lines
align comments in statements
Style functionreferences


KevinEady (approved):
🎉

PR #699, opened 2024-08-22
Add support for class instances in Executor

KevinEady:
### Introduce runtime `BClassInstance`
- Stores its index into the program's class descriptor table
- Given a method name, can create a `BFunctionRef` if the class descriptor has that method.
- Inherits from `BStruct`, so behaves like a struct with members. However, it is not serializable (packs to `u`) and does not enumerate its members when stringified (overrides bstruct's `getStringRep()`)
- Class methods are cached in the Executor as a tuple-type `(program, class index, method name)` ... We need to track the program since it can change now (thanks to "allow funcref calls across executors" feature)

### Introduce `BConstObject`
- A subclass of `BObject`, represents an object that cannot be re-assigned in Escript.
- The class instance object passed to a constructor is a BConstObject, fixing the two issues with assignment to `this`: passing to `super()` and returning from ctor.

### Introduce `ins_check_mro`
- Borrowing the term "MRO" (Method Resolution Order) from Python, this instruction skips the next two instructions (`makelocal`, `jump`) if the target of the jump has already executed, as tracked by the  BClassInstance's set of `constructors_called`.
- The instruction takes the `this` parameter location as its instruction offset. If the function is skipped, all arguments passed to the function are erased from the ValueStack.

### Changes to AST Node `ClassInstance`
- Similarly to the executor bimp, the AST node also tracks the class instance descriptor. Needed when calling its constructor, to generate the `ins_classinst` with the correct index.

### Misc
- Adjusted function resolver to register class methods when its class is registered. (see comments for reasoning)
- Removed unused member `ClassDeclaration::super`


pol-core/bscript/bclassinstance.cpp
```diff
@@ -0,0 +1,120 @@
...
+BClassInstance::BClassInstance( const BClassInstance& B ) : BStruct( B, OTClassInstance )
+{
+  prog_ = B.prog_;
+  index_ = B.index_;
+  globals = B.globals;
+}
+
+size_t BClassInstance::sizeEstimate() const
+{
+  return sizeof( BClassInstance ) + prog_->sizeEstimate();
```
turleypol, line 24: I don't think you have to add the complete prog here. That would be only true if the class is the last one holding the program

turleypol (approved):

pol-core/bscript/bclassinstance.cpp
KevinEady, line 24:
```suggestion
  return sizeof( BClassInstance );
```

turleypol (approved):

PR #701, opened 2024-08-25
Add `classinst.function` member; Add `is` binary operator; Add funcref `MTH_NEW`

KevinEady:
### Introduce `is` operator
- `IS` has been a reserved word since The Beginning ™, so it's been implemented for the `classInstance is funcRef` syntax, which returns `true` if `funcRef` is in the list of constructor functions for `classInstance`
- Added class machinery functions to BObjectImp to support `is` operator, which is currently only implemented by `BFunctionRef`.

### Introduce `.function`, `MBR_FUNCTION` as a navigation suffix
- Used in class instances, accessing `MBR_FUNCTION` will return the _first_ constructor in the list of constructors for `classInstance`.

### Introduce `.new`, `MTH_NEW` to function references
- Allows creation of new objects at runtime (vs. compile-time)
- This makes the equivalent to calling `Foo()` to construct vs. `@Foo.new()` (as `@Foo.call()` expects a `this`)

### `ins_check_mro` Changes
- Removed fatal error if instruction executed on non-classinst. Should not be a _fatal_ error if doing eg. `@Foo.call( struct{} )` -> the `super()` just won't be called.

### Program table changes
- Added the PC/address to the Function Reference table, and it is no longer part of `ins_funcref` instruction.
- This simplified a lot of places where we were tracking both a PC/address as well as funcref index.
- Changed the constructor list in the Class Descriptor Table to be funcref indices instead of PC/addresses

### Misc changes
- Fixed issue with listfile generation, where `get-method-id` listfile instruction was printing with the method id of `ins.type` instead of the correct `ins.lval`


pol-core/bscript/eprog.h
```diff
@@ -80,31 +80,34 @@ struct EPExportedFunction
...
+    return address == other.address && parameter_count == other.parameter_count &&
+           capture_count == other.capture_count && is_variadic == other.is_variadic &&
+           class_index == other.class_index;
   }
 };
 
 struct EPMethodDescriptor
 {
-  unsigned address;
+  // unsigned address;
```
KevinEady, line 99: remove comment

pol-core/bscript/object.cpp
```diff
@@ -2418,6 +2479,20 @@ BObjectImp* BFunctionRef::call_method_id( const int id, Executor& ex, bool /*for
                                       ex.numParams() ) );
     }
   }
+  case MTH_NEW:
+  {
+    if ( variadic_ )
+    {
+      return new BError( fmt::format( "Invalid argument count: expected {}+, got {}",
+                                      static_cast<int>( num_params_ ) - 2 /*remove this*/,
```
KevinEady, line 2487: no need for cast, and probably quote "this" so it's not "remove this" (which may imply "remove this" the code)

turleypol (approved):

PR #702, opened 2024-08-27
ssopt.MobilesBlockNpcMovement allows followers to stack on a tile

Reloecc:
ssopt.MobilesBlockNpcMovement now accepts values 0/1/2.
Earlier 0 = false, 1 = true.
Now 2 = npcs of the same master are allowed to walk on top of each other.
Backward compatible.


KevinEady:
Hi @Reloecc ,

We understand that there are other ssopt settings that are ternary 0/1/2 (`InvulTag`, `SupportFaces`) and the code just refers to their values as `0`, `1`, and `2`, but these "magic values" are tricky to understand in the code. 

Could you define an enum for this option in ssopt.h and use this for the 0/1/2 checks in the code? Would be greatly appreciated!

Reloecc:
I understand the need of an Enum. Commited the change, thank you. If anything goes wrong in the code, please, let me know.

turleypol (approved):

turleypol (approved):

PR #703, opened 2024-08-27
Generate constructors for child classes

KevinEady:
### Second-pass AST Generation Changes
- Uses polymorphism on `AvailableParseTree` (now renamed to `SecondPassTarget`) to split into AST building targets into ANTLR nodes (eg. building a user-defined function AST node via a `AvailableParseTree`) and generated functions (eg.  building a super() function via `AvailableGeneratedFunction`)
- Make generated functions have scoped function parameters, used to prevent duplicate function parameters being generated for super().
  - See PR comments for example of compiler-generated ctors and supers

### DIFF
--- a/pol-core/bscript/compiler/astbuilder/GeneratedFunctionBuilder.cpp
+++ b/pol-core/bscript/compiler/astbuilder/GeneratedFunctionBuilder.cpp
@@ -11,6 +11,7 @@
 #include "bscript/compiler/ast/FunctionCall.h"
 #include "bscript/compiler/ast/FunctionParameterDeclaration.h"
 #include "bscript/compiler/ast/FunctionParameterList.h"
+#include "bscript/compiler/ast/FunctionReference.h"
 #include "bscript/compiler/ast/GeneratedFunction.h"
 #include "bscript/compiler/ast/Identifier.h"
 #include "bscript/compiler/ast/ReturnStatement.h"
@@ -61,6 +62,8 @@ void GeneratedFunctionBuilder::super_function( std::unique_ptr<GeneratedFunction
           base_class_ctors.push_back( base_class_ctor );
         }
       }
+      to_link.insert( to_link.end(), base_cd->base_class_links.begin(),
+                      base_cd->base_class_links.end() );
     }
   }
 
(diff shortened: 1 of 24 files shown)


KevinEady:
The "diamond problem" `A - B&C - D`, where B&C do not have constructors.

For compiler-generated constructors, the parameter will be scoped to the baseclass ctor, eg. we see here for the generated `B` and `C`, their first parameter is a scoped `A::arg0`
```
Ctor B function: - generated-function(B)
  - function-parameter-list
    - function-parameter-declaration(this, byref)
    - function-parameter-declaration(A::arg0)
  - user-function-body
    - value-consumer
      - function-call(B::B)
        - argument()
          - identifier(this)
        - argument()
          - identifier(A::arg0)
```
```
Ctor C function: - generated-function(C)
  - function-parameter-list
    - function-parameter-declaration(this, byref)
    - function-parameter-declaration(A::arg0)
  - user-function-body
    - value-consumer
      - function-call(C::C)
        - argument()
          - identifier(this)
        - argument()
          - identifier(A::arg0)
```
The super function generates its function parameters _also_ based on the scoped parameter names, and will skip parameters that are duplicate. For example, `D`s super function only takes _one_ `A::arg0` even though both `B` and `C` require an argument: since they are both `A::arg0`, only one is needed. We see the func parameters only has one arg0, but both intermediate parent class ctors `B` and `C` are called with this arg:
```
Super function: - generated-function(super)
  - function-parameter-list
    - function-parameter-declaration(this, byref)
    - function-parameter-declaration(A::arg0)
  - user-function-body
    - value-consumer
      - function-call(C::C)
        - argument(this)
          - identifier(this)
        - argument(A::arg0)
          - identifier(A::arg0)
    - value-consumer
      - function-call(B::B)
        - argument(this)
          - identifier(this)
        - argument(A::arg0)
          - identifier(A::arg0)
    - return-statement
      - identifier(this)
```

pol-core/bscript/compiler/astbuilder/UserFunctionBuilder.cpp
```diff
@@ -204,12 +221,12 @@ std::unique_ptr<UserFunction> UserFunctionBuilder::make_user_function(
       bool first = !class_name.empty();
       for ( auto param : param_list->functionParameter() )
       {
-        ScopableName parameter_name( ScopeName::None, text( param->IDENTIFIER() ) );
+        auto param_text = text( param->IDENTIFIER() );
```
KevinEady, line 224: These changes can be reverted

pol-core/bscript/compiler/astbuilder/GeneratedFunctionBuilder.cpp
```diff
@@ -31,44 +34,110 @@ GeneratedFunctionBuilder::GeneratedFunctionBuilder( const SourceFileIdentifier&
...
+          fmt::format( "class {} has linked ctor", class_declaration->name ) );
+
+    class_declaration->internal_error(
+        fmt::format( "class {} unlinked ctor", class_declaration->name ) );
+  }
+
+  to_visit.insert( to_visit.end(), class_declaration->base_class_links.begin(),
+                   class_declaration->base_class_links.end() );
+
+  // while ( !to_visit.empty() )
```
KevinEady, line 96:
```suggestion
```

PR #705, opened 2024-08-28
Add semantic error on cyclic inheritance

KevinEady:
### Semantic Analysis changes
- Error if a child class references itself through cyclic inheritance, eg. `A` inherits `B`, `B` inherits `A`.

### DIFF
--- a/testsuite/escript/classes/fail-self-base-inheritance.err
+++ b/testsuite/escript/classes/fail-self-base-inheritance.err
@@ -1,0 +1,1 @@
+fail-self-base-inheritance.src:10:1: error: Class 'Animal' references itself as a base class through inheritance.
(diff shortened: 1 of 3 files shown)


testsuite/escript/classes/fail-self-base-inheritance.err
```diff
@@ -0,0 +1 @@
+fail-self-base-inheritance.src:10:1: error: Class 'Animal' references itself as a base class through inheritance.
```
KevinEady, line 1: There are actually _three_ errors: each class references itself, but we can't really check multiple errors in our testsuite currrently:
```
fail-self-base-inheritance.src:10:1: error: Class 'Animal' references itself as a base class through inheritance.
fail-self-base-inheritance.src:7:1: error: Class 'BaseClass2' references itself as a base class through inheritance.
fail-self-base-inheritance.src:3:1: error: Class 'BaseClass1' references itself as a base class through inheritance.
```

turleypol (approved):

PR #710, opened 2024-09-01
Support function reference calls with default parameters

KevinEady:
### Function Reference objimp and table changes
- Remove duplication of members across `BFunctionReference` and the function reference table
  - `BFunctionReference` now only contains its index into the function reference table
- Add `is_constructor` to function reference table, which is checked during `MTH_NEW` calls
- Add addresses/PCs of default parameters (see below)

### Refactor handling of default arguments
- Emit instructions for default arguments after visiting all user functions
- `BFunctionRef::validCall()`, which sets the `Instruction jmp` address to the user function's address: now will jump to the instruction corresponding to the missing default argument's addition to the `ValueStack`. 
  - The final instruction in the list of default arguments is a `jump <user function address>`, going to the correct user function location.
- Introduce `BFunctionReference::expected_args()`, returning a pair<int,int> representing the minimum and maximum expected arguments for the function (_without_ considering variadic args)
- Use this `expected_args()` when validating calls, generating error messages
  - Additionally, simplify the logic a bit by always constructing `new BClassInstance()` for `MTH_NEW` or the existing instance for `MTH_CALL`. This removes that `add_new_classinst` parameter from `call_function_reference`. 
- Add `MTH_CALL_METHOD`, needed _specifically_ for creating a correct `BError` message for invalid parameter count
  - Eg. a script call of `obj.method()` (zero args) gets runtime-executed as the equivalent of `method(obj)` (one arg), so we need subtract one for the user-shown error message.

### Misc
- Introduce `const` variant of `Node::child(size_t index)`
- The CodeGenerator was erroneously creating a new `user_function_labels` map, when the CompilerWorkspace already has one.
- Remove `visit_class_declaration` from InstructionGenerator, as it does not _generate_ any _instructions_.
  - It is now in the caller, `CodeGenerator::generate_instructions`


pol-core/bscript/compiler/codegen/CodeGenerator.cpp
```diff
@@ -118,9 +114,20 @@ void CodeGenerator::generate_instructions( CompilerWorkspace& workspace )
     }
   }
 
+  // The default parameter generation needs to happen after generating the
+  // instructions for the user functions, as only class method functions and
+  // user functions with a registered function reference emit their default
+  // arguments. Since a user function may be registered as a function reference
+  // only _during_ some other user function's generation, we are only sure a
+  // user function has a function reference after generating all of them.
+  for ( const auto& user_function : workspace.user_functions )
```
KevinEady, line 123: This adds another `O(n)` loop at the benefit of having smaller ecl program size. 

In this implementation, a script with 100 user functions with only one as a function reference will emit default argument instructions for that _one_ function (but still loops through all 100 functions to check if it's referenced).

We can remove this second-pass loop by emitting the default arguments for _all_ user functions, but then we could have a gigantic ECL, as all user functions -- regardless of `@FuncRef` usage -- would have their default arguments emitted.

pol-core/bscript/executor.cpp
```diff
@@ -3729,28 +3760,16 @@ void Executor::call_function_reference( BFunctionRef* funcr, BContinuation* cont
...
-      return;
-    }
-
-    fparams.insert( fparams.begin(), BObjectRef( new BConstObject( new BClassInstance(
-                                         prog_, funcr->class_index(), Globals2 ) ) ) );
-  }
+  auto nparams = static_cast<int>( fparams.size() );
 
-  if ( funcr->variadic() )
+  // Do we need to create an array{} for a variadic function call?
```
KevinEady, line 3765: Could use a better comment, this sounds like a leftover todo when its not

turleypol (approved):

Issue #545, opened 2023-09-23
Configurable View Range


Zrnik:
Hello, i have a request for a change that would make my life much easier, i have to build my own version of pol core because i want my client to allow 40 tiles visual range, which i have to change on those lines:

https://github.com/polserver/polserver/blob/eaa96e62dd25fa0df3b59c281150e34e986f80e5/pol-core/plib/uconst.h#L53
https://github.com/polserver/polserver/blob/eaa96e62dd25fa0df3b59c281150e34e986f80e5/pol-core/plib/uconst.h#L54

(and here for tests https://github.com/polserver/polserver/blob/eaa96e62dd25fa0df3b59c281150e34e986f80e5/testsuite/pol/testpkgs/house/test_houserange.src#L41 )

I am not very much into c++ ecosystem and compiling it by my self is pain (+ i want up 2 date core, so i have to merge your changes once a while)

This is actually the only change i have in my custom built core. 

I am not capable to achieve this myself in c++ (otherwise i would submit a pull request) so is there a posibility, that you would add configuration entry in `pol.cfg` for example that will set this visual range? 

`MaxClientVisualRange 40` for example?

Im willing to pay for this change.

Pumpk1ns:
I like that idea. But are you sure that it doesn't cause strange side effects?

Zrnik:
It would in regular client, I have also modified classicuo client that was updated to support the extended view range

PR #715, opened 2024-09-16
Configurable visual range

Kukkimonsuta:
This PR adds new setting to pol.cfg "VisualRange". This is a constant converted to configurable value which changes the distance at which client can see its surroundings. Changing the value requires changing the client as well. Default value is kept at 18 (=unchanged). Do note that this PR doesn't touch on the "client reported visual range" at all which seems to be present in an unfinished state.

The bad thing is I'm not sure how to properly test this as misconfiguration can have weird consequences and this PR heavily depends on the fact the constant is being used everywhere it needs to.

The good thing is keeping the configured value at 18 should keep the current behavior.

Fixes: https://github.com/polserver/polserver/issues/545

### DIFF
--- a/pol-core/pol/globals/uvars.cpp
+++ b/pol-core/pol/globals/uvars.cpp
@@ -171,8 +171,8 @@ GameState::GameState()
       uo_skills(),
       task_thread_pool(),
       decay(),
-      max_update_range( Plib::RANGE_VISUAL ),
-      max_update_range_client( Plib::RANGE_VISUAL ),
+      max_update_range( 0 ),
+      max_update_range_client( 0 ),
       max_update_range_multi( 0 )
 
 {
@@ -185,7 +185,11 @@ GameState::~GameState()
   // or make sure that the globals get deconstructed before eg the flyweight string container
 }
 
-
+void GameState::initialize_range_from_config( u16 range )
+{
+  max_update_range_client = range;
+  max_update_range = max_update_range_multi + max_update_range_client;
+}
 void GameState::update_range_from_multis()
 {
   for ( const auto& m_pair : Multi::multidef_buffer.multidefs_by_multiid )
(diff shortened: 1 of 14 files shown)


Kukkimonsuta:
https://github.com/polserver/polserver/pull/715/commits/959c212c6382019f3f386b17100e3f4a47cacc46

Not sure what type is correct here, it's supposed to be byte, so all of `u8` `std::uint8_t` or simply `unsigned char` would work, but `u8` failed on ci for some reason (assuming missing header, no idea why it works locally though). Anyway let me know if `std::uint8_t` is fine or want it changed.

Kukkimonsuta:
rebased onto latest master

pol-core/pol/globals/uvars.cpp
```diff
@@ -185,7 +185,11 @@ GameState::~GameState()
   // or make sure that the globals get deconstructed before eg the flyweight string container
 }
 
-
+void GameState::initialize_range_from_config( u16 range )
+{
+  max_update_range_client = range;
```
turleypol, line 190: Since pol.cfg can be reread during runtime the new value should be only applied if it's greater then the current `max_update_range_client`

Kukkimonsuta: At the moment this is called only on initial config load and not on reload ( https://github.com/polserver/polserver/blob/fb034e0d113acdbf8f74fb7e8fb44c9db0148fb5/pol-core/pol/polcfg.cpp#L55-L115 )

I'm not sure whether allowing reload is something we want to do because:
1) it needs to be in sync with client, I do not expect people changing this often
2) the dynamic view range doesn't seem to be properly implemented, so i'm bit worried about what it actually does https://github.com/polserver/polserver/blob/f45d12665d5b710f1643c96b44faaf1219c972ba/pol-core/pol/miscmsg.cpp#L475  https://github.com/polserver/polserver/blob/f45d12665d5b710f1643c96b44faaf1219c972ba/pol-core/pol/mobile/charactr.cpp#L4319

That being said I can change it if you confirm these aren't issues.

turleypol (approved):

turleypol (approved):

PR #720, opened 2024-10-13
Removed PolLock when handling nonallowed pkts

turleypol:
Currently when a undefined pkt got send the login/client thread does not take PolLock, but if it's a defined one it always takes the lock. For non-allowed pkts this is not needed.

For debugging locks added a define in polsem.h to log each usage.


KevinEady (dismissed):
LGTM!

KevinEady (approved):

PR #721, opened 2024-10-15
Socket::has_incomming_data closes socket on error

turleypol:
eg auxclient via OpenConnection had the problem that a closed connection by the other side wasnt detected.
cleanup timeout params uses now milliseconds
removed old unused pollingwithselect
renamed pol.cfg entry for SingleThreadLoginServer socket timeout


KevinEady (approved):
LGTM!

PR #726, opened 2024-10-19
added ssopt AllowVisualRangeModification

turleypol:
Which allows clients to modify their LOS via pkt 0xC8


pol-core/pol/miscmsg.cpp
```diff
@@ -467,12 +467,14 @@ void handle_unknown_C4( Client* client, PKTOUT_C4* /*msg*/ )
 void handle_update_range_change( Client* client, PKTBI_C8* msg )
 {
   // limit range to official range 5-24
-  u8 range = std::clamp( msg->range, (u8)5, (u8)24 );
-  client->set_update_range( range );
+  if ( settingsManager.ssopt.allow_visual_range_modification )
+  {
+    u8 range = std::clamp( msg->range, (u8)5, (u8)24 );
```
Kukkimonsuta, line 472: I don't want to be annoying, but could this be a configuration as well - something like `MinimumVisualRange` / `MaximumVisualRange`? If we were to allow clients changing it by themselves on our shard we'd most likely want to allow only to lower it.

Kukkimonsuta:
Thinking about this bit more - how difficult would it be to also add override at runtime - ex. no matter what client or default says, view range is X for the time the property is set? For instance imagine we'd make "very dark" dungeon where you can see only few tiles around 👀 With original client it would probs look not so good, but with classicuo we've got plenty of options how to make it work 😄

pol-core/pol/uoscrobj.cpp
```diff
@@ -4930,6 +4932,10 @@ BObjectImp* Client::set_script_member_id( const int id, int value )
   case MBR_DISABLE_INACTIVITY_TIMEOUT:
     disable_inactivity_timeout( value );
     return new BLong( disable_inactivity_timeout() );
+  case MBR_VISUAL_RANGE:
+    set_update_range( Clib::clamp_convert<u8>( value ) );
+    gd->script_defined_update_range = true;
```
Kukkimonsuta, line 4937: Would it be reasonable to allow setting to `uninit` to reset to whatever user had set before/default value and allow client to change it again?

Kukkimonsuta:
When reducing the range it seems to work nicely.

When increasing the range it doesn't show items that were out of range before - should client request resync in that case? Or should they be sent automatically?

pol-core/pol/miscmsg.cpp
turleypol: Added

turleypol:
> Thinking about this bit more - how difficult would it be to also add override at runtime - ex. no matter what client or default says, view range is X for the time the property is set? For instance imagine we'd make "very dark" dungeon where you can see only few tiles around 👀 With original client it would probs look not so good, but with classicuo we've got plenty of options how to make it work 😄

Added escript member for that which now also removes objects and adds objects

turleypol:
> Tested both server-controlled and client-controlled.
> 
> When reducing the range it seems to work nicely.
> 
> When increasing the range it doesn't show items that were out of range before - should client request resync in that case? Or should they be sent automatically?

Should be fixed now

Kukkimonsuta:
Did bunch of tests and all seems good. Only feedback i have is whether it's really required to remove everything upon changing the range? It causes somewhat nasty flickering, but it's def something one could live with.

https://github.com/user-attachments/assets/060a48cf-6682-4811-a88f-f6834748f583

Also I cannot verify how original client works, however default classicuo macros for changing view distance will just do it locally and won't send any packets to server to let it know what range is set - our classicuo is modified to send 0xC8 to server and wait for reply to actually change it. I don't think there is much we can do about that as by default classicuo sends 0xC8 only after login.

pol-core/pol/miscmsg.cpp
KevinEady: Resolving as this was addressed

pol-core/pol/uoscrobj.cpp
KevinEady: Resolving as this was addressed: setting to 0

KevinEady (approved):
LGTM!

PR #729, opened 2024-10-20
Add moved_at and position_changed_at members to characters

Kukkimonsuta:
This PR adds new read-only character members:

- `moved_at` - last time at which movement was made (meaning character actually moved due to move function)
- `position_chaged_at` - last time at which character changed position (this includes things like teleports or `MoveObjectToLocation` atc.)

This is needed for mechanics that depend on "can't use while moving" with various accuracy (ex. some may require player to stand still for a second, some just for half etc.).

### DIFF
--- a/docs/docs.polserver.com/pol100/objref.xml
+++ b/docs/docs.polserver.com/pol100/objref.xml
@@ -309,8 +309,8 @@
 <member mname="luck_mod" type="Integer" access="r/w">Modifier for property, added to base property.</member>
 <member mname="swing_speed_increase_mod" type="Integer" access="r/w">Modifier for property, added to base property.</member>
 <member mname="buffs" type="dictionary" access="r/o">Returns the current active buffs added via addBuff as dictionary. Key is the iconid and value a struct struct{name_cliloc, desc_cliloc, end_time, name_args, desc_args}.</member>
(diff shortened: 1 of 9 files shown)


docs/docs.polserver.com/pol100/objref.xml
```diff
@@ -309,6 +309,8 @@
 <member mname="luck_mod" type="Integer" access="r/w">Modifier for property, added to base property.</member>
 <member mname="swing_speed_increase_mod" type="Integer" access="r/w">Modifier for property, added to base property.</member>
 <member mname="buffs" type="dictionary" access="r/o">Returns the current active buffs added via addBuff as dictionary. Key is the iconid and value a struct struct{name_cliloc, desc_cliloc, end_time, name_args, desc_args}.</member>
+<member mname="position_changed_at" type="Integer" access="r/o">PolClock when character last changed position for any reason</member>
```
turleypol, line 312: Instead PolClock I would suggest to mention the MilliSecondClock which you use to store the value. And you return a Double and not a Integer.

turleypol: Double is correct otherwise it will overflow after a few days.
The others return in 10ms resolution, not really usable I guess, but it is what it is. :)

turleypol (approved):

PR #734, opened 2024-11-07
Add realm decay setting

Kukkimonsuta:
As discussed on discord, this PR allows opting out of decay for a whole realm.

The PR is still missing documentation, however code is ready for review.

New script methods:
```
GetRealmDecay( realm );
SetRealmDecay( realm, has_decay );
```

The changes include slight modification on how decay is handled - before each step would always clear an area, however now a step can be just moving onto next realm by itself. This was changed because I wanted to avoid possible infinite loop when all realms have decay disabled as well as allow for bit more comfortable unit testing.

### DIFF
--- a/pol-core/pol/module/polsystemmod.cpp
+++ b/pol-core/pol/module/polsystemmod.cpp
@@ -407,6 +407,37 @@ BObjectImp* PolSystemExecutorModule::mf_DeleteRealm( /*name*/ )
   return new BLong( 1 );
 }
 
+BObjectImp* PolSystemExecutorModule::mf_GetRealmDecay( /*name*/ )
+{
+  const String* realm_name;
+  if ( !getStringParam( 0, realm_name ) )
+  {
+    return new BError( "Invalid parameter" );
+  }
+  Realms::Realm* realm = Core::find_realm( realm_name->value() );
+  if ( !realm )
+    return new BError( "Realm not found." );
+
+  return new BBoolean( realm->has_decay );
+}
+
+BObjectImp* PolSystemExecutorModule::mf_SetRealmDecay( /*name,has_decay*/ )
+{
+  const String* realm_name;
+  bool has_deacy;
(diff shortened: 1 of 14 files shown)


pol-core/pol/module/polsystemmod.cpp
```diff
@@ -407,6 +407,37 @@ BObjectImp* PolSystemExecutorModule::mf_DeleteRealm( /*name*/ )
...
+  bool has_deacy;
+  if ( !( getStringParam( 0, realm_name ) && getParam( 1, has_deacy ) ) )
+  {
+    return new BError( "Invalid parameter" );
+  }
+  Realms::Realm* realm = Core::find_realm( realm_name->value() );
+  if ( !realm )
+    return new BError( "Realm not found." );
+
+  realm->has_decay = has_deacy;
```
turleypol, line 436: Call of Decay realm_size_change to calc the new sleep time?

pol-core/pol/savedata.cpp
```diff
@@ -209,13 +209,18 @@ void write_shadow_realms( Clib::StreamWriter& sw )
 {
   for ( const auto& realm : gamestate.Realms )
   {
-    if ( realm->is_shadowrealm )
+    sw.begin( "Realm" );
```
turleypol, line 212: Please change the name of the method like you did for reading

pol-core/pol/uimport.cpp
```diff
@@ -279,18 +279,29 @@ void read_system_vars( Clib::ConfigElem& elem )
...
+      add_realm( name, baserealm, has_decay );
+      INFO_PRINTLN( "\nShadowrealm {}", name );
+    }
+  }
   else
   {
-    add_realm( name, baserealm );
-    INFO_PRINTLN( "\nShadowrealm {}", name );
+    // these are dynamic settings for base realm
+    Realms::Realm* realm = find_realm( name );
```
turleypol, line 303: The realm can no longer exists, but I would say no error just a warning.

turleypol (approved):

PR #735, opened 2024-11-11
added multi prop/itemdesc items_decay

turleypol:
added region cfg entry NoDecay
misc cleanup eg chr->equip sets position instead of calling side


KevinEady (dismissed):

KevinEady (approved):
LGTM!

PR #743, opened 2025-01-13
remove GIT_SHALLOW from external cmake projects

KevinEady:


KevinEady:
Closed in favor of #742

PR #754, opened 2025-01-30
Add missing attack range properties to itemdesc

AsYlum-:


pol-core/pol/item/itemdesc.cpp
```diff
@@ -726,6 +726,10 @@ ItemDesc::ItemDesc( Type type )
       luck_mod( 0 ),
       swing_speed_increase( 0 ),
       swing_speed_increase_mod( 0 ),
+      min_attack_range_increase( 0 ),
+      min_attack_range_increase_mod( 0 ),
```
turleypol, line 730: Something like:
`Creator for empty itemdesc, used when no entry exists
thus all members need to be meaningful initialized!` as comment at the beginning of the constructor?
And in line 244 `do not forget the other constructor!`

turleypol (approved):

Issue #749, opened 2025-01-23
Packet 0xF3 sending only Movable flag.


daedel:
Hi, I can see that in documentation of packet 0xf3 there is information that flags are sends for item information.
But currently I can see that only movable flag is acctualy sending.

<img width="561" alt="Image" src="https://github.com/user-attachments/assets/0a24de51-2bef-4146-9d96-6b5ee3110fce" />

Can you fix that to send also flags that item is hidden when player can see invisible items?

KevinEady:
Hi @daedel ,

Thanks for reporting! We've fixed this issue in #755. It will be available in all nightly builds that include this PR. 

Kevin

PR #755, opened 2025-02-01
Add hidden flag to invisible items

KevinEady:
Adds the `HIDDEN` flag to items, so clients will show them as hidden, eg:
image

Fixes: #749

### DIFF
--- a/testsuite/pol/testpkgs/client/test_client_seeinvisitems.src
+++ b/testsuite/pol/testpkgs/client/test_client_seeinvisitems.src
@@ -1,0 +1,85 @@
+include "communication";
+include "testutil";
+
+use uo;
+
+const ITEM_FLAG_HIDDEN := 0x80;
+
+var char;
+var charX := 100;
+var charY := 50;
+var clientcon := getClientConnection();
+
+program test_item_move()
+  var a := FindAccount( "testclient0" );
+  char := a.getcharacter( 1 );
+  if ( !char )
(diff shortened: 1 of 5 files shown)


testsuite/pol/testpkgs/client/test_client_seeinvisitems.src
```diff
@@ -0,0 +1,85 @@
...
+    ev := waitForClient( 0, { EVT_NEW_ITEM } );
+    if ( !ev )
+      char.disable( "seeinvisitems" );
+      return ev;
+    elseif ( ev.serial == item.serial )
+      break;
+    endif
+  endwhile
+
+  if ( ev.status != ITEM_FLAG_HIDDEN )
```
frozenblit, line 78: can the test get a bit flaky in case the item gets a movable flag later and instead do an ( ... & ITEM_FLAG_HIDDEN) != ITEM_FLAG_HIDDEN?

frozenblit (approved):

testsuite/pol/testpkgs/client/test_client_seeinvisitems.src
KevinEady: Addressed in https://github.com/polserver/polserver/pull/755/commits/ace9e0c137e21d5c25b4a7cac54fcb07eed363f4

frozenblit (approved):

PR #758, opened 2025-02-06
Worldsave further improvements

turleypol:
Directly use the c API for writing savefiles to get rid of unneeded overhead.
Increased local buffer to 32kb per file and write without extra buffer to the files.

Seems to give us 20% faster write times.

### DIFF
--- a/pol-core/clib/streamsaver.cpp
+++ b/pol-core/clib/streamsaver.cpp
@@ -2,20 +2,26 @@
 #include <fstream>
 #include <ios>
 #include <iostream>
+#include <stdexcept>
 #include <string>
 
 #include "streamsaver.h"
 
 namespace Pol::Clib
 {
-StreamWriter::StreamWriter( std::ostream& stream ) : _stream( stream ) {}
+StreamWriter::StreamWriter( const std::string& path ) : _file( fopen( path.c_str(), "wb+" ) )
+{
+  if ( !_file )
+    throw std::runtime_error{ fmt::format( "failed to open {}", path ) };
+  setbuf( _file, nullptr );  // disable buffer
+}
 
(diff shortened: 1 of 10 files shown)


pol-core/clib/streamsaver.cpp
```diff
@@ -2,20 +2,26 @@
...
 
 #include "streamsaver.h"
 
 namespace Pol::Clib
 {
-StreamWriter::StreamWriter( std::ostream& stream ) : _stream( stream ) {}
+StreamWriter::StreamWriter( const std::string& path ) : _file( fopen( path.c_str(), "wb+" ) )
+{
+  if ( !_file )
+    throw std::runtime_error{ fmt::format( "failed to open {}", path ) };
```
KevinEady, line 15: **question**: Will this throw be correctly caught by your recent work with exception handling and saving? ie, this will _not_ be a silent error?

KevinEady (approved):
LGTM, just a q

pol-core/clib/streamsaver.cpp
turleypol: This exception will be caught here: https://github.com/polserver/polserver/blob/c588c51a33ae2b891a36237f173652a935aab2db/pol-core/pol/savedata.cpp#L503
(The exceptions during writing will be caught in the save lambda which set the same promise)
Sets the promise to false. Which the uomod function waits for and then returns false:
https://github.com/polserver/polserver/blob/c588c51a33ae2b891a36237f173652a935aab2db/pol-core/pol/savedata.cpp#L519
And finally:
https://github.com/polserver/polserver/blob/c588c51a33ae2b891a36237f173652a935aab2db/pol-core/pol/module/uomod.cpp#L3077

PR #778, opened 2025-05-06
Allow large custom housing by sending floor as multiple planes

KevinEady:


KevinEady:
In draft because currently based on https://github.com/nodejs/node-addon-api/pull/1739 to test the ICE failure.

Will rebase to `main` once https://github.com/nodejs/node-addon-api/pull/1739 is merged.

frozenblit (approved):

PR #796, opened 2025-07-18
Add ecompile flag to print AST

KevinEady:


pol-core/bscript/compiler/codegen/CodeGenerator.cpp
```diff
@@ -73,11 +74,20 @@ std::unique_ptr<CompiledScript> CodeGenerator::generate(
 
   std::vector<ClassDescriptor> class_descriptors = class_declaration_registrar.take_descriptors();
 
+  std::string tree;
+
+  if ( generate_ast_string )
+  {
+    AbstractSyntaxTreeStringGenerator generator;
```
turleypol, line 81: Windows warning: `declaration of 'generator' hides previous local declaration` :)

KevinEady: Addressed in https://github.com/polserver/polserver/pull/796/commits/ac85e20c2e2ba5c20e19d0910d00092bab74a353

PR #797, opened 2025-07-20
Short Circuit Optimizer

turleypol:


KevinEady (approved):
Just some non-blocking nitpicks, LGTM otherwise

PR #799, opened 2025-07-29
use c++20

turleypol:
* increased used clang version
* compiler report and logfacility use now compile time formatting, which means that the formatstring gets checked at compile time. (which found 2 errors) adapted a few places since report only accepts formatting arguments adapted a few places with logging since char[] is compile time formatting and string or chr* is runtime formatting
* use std::ranges instead of boost
* disabled pragma_assume vs specific macro (I guess noone cares)
* needed to fix ancient ms exception code
* modernized SpinLock
* removed unused code in ECompile
* replaced std::filesystem::path::u8string with string. It now returns an actual u8string type
* cleanup layers.h added the defines for other layers which where before only defined in the pkt
* osmod::OpenConnection and HTTPRequest cleanup: early outs, dont check for pChild which is only needed for startscript and placed suspend at the very last position


pol-core/clib/StdAfx.h
```diff
@@ -19,6 +19,7 @@
 #include <memory>
 #include <mutex>
 #include <queue>
+#include <ranges>
```
KevinEady, line 22: **question**: why is this include needed...?

pol-core/clib/compilerspecifics.h
```diff
@@ -8,13 +8,13 @@ Remove the include in all StdAfx.h files or live with the consequences :)
...
-#define passert_assume( x ) __analysis_assume( x )
-#else
+// #ifdef _MSC_VER
+// #pragma warning( disable : 4467 )  // ATL attributes
+// #include <CodeAnalysis/sourceannotations.h>
+// #define passert_assume( x ) __analysis_assume( x )
+// #else
 #define passert_assume( x )
-#endif
+// #endif
```
KevinEady, line 17: **nitpick**: can remove these comments

KevinEady (dismissed):

pol-core/clib/StdAfx.h
turleypol: Since ranges is likely be used quite often it could benefit from being in our precompiled header

PR #818, opened 2025-09-22
Add support for regular expressions

KevinEady:
- Add regular expressions through `basic::RegExp( expr, flags := "" )
- Add String method `replace( RegExp pattern, FuncRef|String replacement )`
- Add String method `match( RegExp pattern )`
- Add overload for String method `find( RegExp pattern[, Long start] )`


frozenblit (dismissed):

pol-core/bscript/executor.inl.h
```diff
@@ -8,7 +8,14 @@ namespace Pol::Bscript
 template <typename Callback>
 struct CallbackData
 {
-  CallbackData( Callback callback ) : callback( callback ) {}
+  CallbackData( Callback&& callback ) : callback( std::move( callback ) ) {}
+
+  CallbackData( const CallbackData& data ) = delete;
+  CallbackData& operator=( const CallbackData& data ) = delete;
+  CallbackData( CallbackData&& data ) = default;
+  CallbackData&& operator=( CallbackData&& data ) = delete;
```
turleypol: Should be default not delete

PR #820, opened 2025-09-27
Move only callback

turleypol:


KevinEady:
Included in https://github.com/polserver/polserver/pull/818 as https://github.com/polserver/polserver/pull/818/commits/a15c4ea9f7c5504a1a09c7dd0f6930a9f1bc0d39

PR #834, opened 2025-12-22
More cmake options and fixed build of more esoteric platforms

turleypol:
Define the path for the binaries and the download/build folder of the external libs
Option to use the header only version of boost (otherwise I simply don't get it to work)
Fixed curl linking on android
Fixed compiler warning with latest clang
Updated efsw lib
Removed ancient libc check during pol startup

### DIFF
--- a/cmake/Efsw.cmake
+++ b/cmake/Efsw.cmake
@@ -1,7 +1,7 @@
 message("* libefsw")
 
 set (EFSW_REPOSITORY "https://github.com/SpartanJ/efsw.git")
-set (EFSW_VERSION_REF "1.5.0")
+set (EFSW_VERSION_REF "bbc804b")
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -66,7 +69,7 @@ include(TestBigEndian)
 include(cmake/init.cmake)
 
 set(POL_EXT_LIB_DIR "${CMAKE_CURRENT_SOURCE_DIR}/lib")
```
frozenblit: was POL_EXT_LIB_DIR left like this on purpose? because it's the same value as in EXT_DOWNLOAD_DIR

frozenblit (approved):

CMakeLists.txt
turleypol: Yes that's still needed for libraries we have integrated in our repo like sql

PR #839, opened 2025-12-31
do not send warmode change to clients who cant see the character

Kukkimonsuta:
When character changes warmode, packet 0x77 is sent to everyone in range no matter whether the clients around can see the character changing warmode. This isn't issue byt itself as client will just ignore the packet, however 3rd party tools or customized clients can be used to detect this change gaining unfair advantage in pvp by discovering that someone is hidden nearby.

I have tried to make a test for this, but failed so this PR is only tested by me manually on cuo + razor.

### DIFF
--- a/docs/docs.polserver.com/pol100/corechanges.xml
+++ b/docs/docs.polserver.com/pol100/corechanges.xml
@@ -2,9 +2,14 @@
 <ESCRIPT>
 	<header>
 		<topic>Latest Core Changes</topic>
-		<datemodified>12-15-2025</datemodified>
+		<datemodified>12-31-2025</datemodified>
 	</header>
 	<version name="POL100.2.0">
+		<entry>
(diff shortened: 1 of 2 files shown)


pol-core/pol/mobile/charactr.cpp
```diff
@@ -3177,6 +3177,9 @@ void Character::set_warmode( bool i_warmode )
             return;
           if ( !chr->in_visual_range( this ) )
             return;
+          if ( !chr->is_visible_to_me( this ) )
```
turleypol: `is_visible_to_me` by default checks also `in_visual_range` so the check before can be removed.

Kukkimonsuta: thanks, fixed

turleypol:
Small core-changes entry would be nice (and please run the build_corechangea script for the online docs)

PR #844, opened 2026-01-03
Integer overflow when using GetConfigInt

### LINKED ISSUE #347: GetConfigString() GetConfigInt() overflow error
### **Issue:**

**GetConfigString()** and **GetConfigInt()** in some certain situations are not working on Linux like it does on Windows.

**GetConfigString()** merely converts the number back into a string.

POL is very helpful and converts any numerical entry in the config.
On Windows, the number is read as 32 bit and overflows - so POL realizes and treats the entry as a string.
On Linux, the number fits in a 64bit integer - but not a signed integer - so it becomes negative.

The culprits are **bobject_from_string** and **convert_numeric (escrutil.cpp)**.

### **How to fix:**

I could think a few ways to go from here:

1. A quick one is to check, when the final number is negative, if the string starts with "-" otherwise assume the same as an overflow.
2. We could also check that the string representation matches the original string (might be a problem with doubles).
3. The third alternative would be a breaking change: convert only when asked for. For instance, when reading with **GetConfigInt()**

-Text quoted from discord chat.

turleypol:
Should fix #347 (other part of the bug was already fixed by #525)

### DIFF
--- a/pol-core/pol/module/cfgmod.cpp
+++ b/pol-core/pol/module/cfgmod.cpp
@@ -154,7 +154,7 @@ Bscript::BObjectRef EConfigElemRefObjImp::OperSubscript( const Bscript::BObject&
   {
     return get_member( str->data() );
   }
-  else if ( const auto* l = obj.impptr_if<const Bscript::BLong>() )
+  if ( const auto* l = obj.impptr_if<const Bscript::BLong>() )
   {
     return get_member( std::to_string( l->value() ).c_str() );
   }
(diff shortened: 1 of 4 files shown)


pol-core/pol/module/cfgmod.cpp
```diff
@@ -578,34 +510,35 @@ Bscript::BObjectImp* ConfigFileExecutorModule::mf_GetConfigIntArray()
...
+      {
+        if ( str->length() >= 1 )
         {
-          Bscript::String* str = static_cast<Bscript::String*>( imp );
-          ar->addElement( new Bscript::BLong( strtoul( str->data(), nullptr, 0 ) ) );
+          try
+          {
+            ar->addElement( new Bscript::BLong( std::stoi( str->value(), nullptr, 0 ) ) );
+          }
+          catch ( const std::logic_error& )
```
frozenblit: might be worth leaving a comment why this catch block is empty

turleypol: Added

frozenblit (approved):

PR #856, opened 2026-01-17
do not include full program in funcref memory estimate

Kukkimonsuta:
While trying to diagnose memory usage we've noticed some crazy numbers in `PolCore().internal(2);` and `PolCore().internal(5);`. I've tracked it down to `FunctionReference` including the program memory estimate in its own which kind of makes sense, but it makes global memory estimates quite useless. For instance our server consumes 9.9 GB in memory, but estimates bloomed up to over 30 GB due to this. After this change the breakdown in `memoryusage.log` sums to 7.92 GB out of 9.91 GB process size, which seems much more reasonable and useful.

### DIFF
--- a/pol-core/bscript/object.cpp
+++ b/pol-core/bscript/object.cpp
@@ -2334,7 +2334,7 @@ BObjectImp* BFunctionRef::copy() const
 
 size_t BFunctionRef::sizeEstimate() const
 {
-  return sizeof( BFunctionRef ) + Clib::memsize( captures ) + prog_->sizeEstimate();
+  return sizeof( BFunctionRef ) + Clib::memsize( captures );
 }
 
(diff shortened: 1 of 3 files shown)


pol-core/bscript/object.cpp
```diff
@@ -2334,7 +2334,7 @@ BObjectImp* BFunctionRef::copy() const
 
 size_t BFunctionRef::sizeEstimate() const
 {
-  return sizeof( BFunctionRef ) + Clib::memsize( captures ) + prog_->sizeEstimate();
+  return sizeof( BFunctionRef ) + Clib::memsize( captures );
```
turleypol: Good catch!
The BFunctionRef also holds `globals` which are missing. 
Could you please add `Clib::memsize( *globals )`

PR #866, opened 2026-02-19
Uoconvert the wrong coordinates where shown when a block contains more then the allowed number of items

turleypol:
Due to wrong calculation to get from Blockindex to coordinates

Can be reproduced with our test map when setting in uoconvert the MaxStaticsPerBlock to eg 40.
The multi is at ~150,150
Without this PR it complains about 2,8 - 9,15
Now it prints 144,144 - 151,151

### DIFF
--- a/pol-core/plib/uofile02.cpp
+++ b/pol-core/plib/uofile02.cpp
@@ -98,7 +98,7 @@ void rawstaticfullread()
         passert_always_r(
             srec_count <= cfg_max_statics_per_block,
             fmt::format(
-                "to many static items in area {},{} - {},{} - maybe double items... you've "
+                "too many static items in area {},{} - {},{} - maybe double items... you've "
                 "to reduce amount of {} items below {} items ",
                 x, y, x + 7, y + 7, srec_count, cfg_max_statics_per_block ) );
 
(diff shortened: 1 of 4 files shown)


pol-core/plib/uofile02.cpp
```diff
@@ -95,15 +94,13 @@ void rawstaticfullread()
...
-                              Clib::tostring( y ) + " " + Clib::tostring( x + 7 ) + " " +
-                              Clib::tostring( y + 7 ) +
-                              " - maybe double items... you've to reduce amount of " +
-                              Clib::tostring( srec_count ) + " items below " +
-                              Clib::tostring( cfg_max_statics_per_block ) + " items" );
+        auto [x, y] = staticblock_to_coords( block, uo_map_height );
+        passert_always_r(
+            srec_count <= cfg_max_statics_per_block,
+            fmt::format(
+                "to many static items in area {},{} - {},{} - maybe double items... you've "
```
KevinEady: **nit**: this "to" should be "too" but it was like that before so i guess it's okay 🙂

pol-core/plib/uofile02.cpp
```diff
@@ -149,23 +146,21 @@ void rawstaticfullread()
...
```
KevinEady: **nit**: "too"

KevinEady (dismissed):

KevinEady (approved):

PR #868, opened 2026-03-07
dont use a static buffer for reading cfgfile to be threadsafe

turleypol:
eg config/www.cfg can load in the background
cleanup the file: deleted ancient define, simplified logic

### DIFF
--- a/pol-core/clib/cfgfile.cpp
+++ b/pol-core/clib/cfgfile.cpp
@@ -24,15 +24,12 @@
 
 namespace Pol::Clib
 {
+using namespace std::literals;
 namespace
 {
 bool commentline( const std::string& str )
 {
-#ifdef __GNUC__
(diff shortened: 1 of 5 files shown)


pol-core/clib/cfgfile.cpp
```diff
@@ -472,144 +390,23 @@ void ConfigFile::open( const char* i_filename )
...
-    {
-      if ( nl != buffer && *( nl - 1 ) == '\r' )
-        --nl;
-      *nl = '\0';
-      strbuf += buffer;
-      return true;
-    }
-
     strbuf += buffer;
+    if ( strbuf.ends_with( '\n' ) )
```
frozenblit, line 408: wasn't the old code stripping the new line as well?

turleypol: Yes correct, but the splitnamevalue function which is used on the returned data also handles newlines. Thus no need to have the additional logic

frozenblit (approved):

PR #884, opened 2026-05-16
Attackable item

turleypol:
* Itemdesc Attackable 1/0
* Sends hp Vs maxhp as statusbar
* Event opponent_moved, engaged, disengaged for controlscript
* Damage against items is calculated like the char base damage:
** Random weapon damage * (tactics+50 + strength * 0.2)* 0.01
* It's always a hit
* No repsys
* Combat hook and hit script are called if defined
* SetOpponent, ListHostiles accepts attackableitems
* ApplyDamage and ApplyRawDamage accept attackableitems, but have no difference
* chr.opponent can return an item ref
* chr.attackonce accept attackable item
* Don't send vital pkts during player/NPC creation it's at that point unknown to the clients


pol-core/pol/item/item.cpp
```diff
@@ -123,6 +126,7 @@ Item* Item::clone() const
   item->name_suffix( name_suffix() );
 
   item->no_drop( no_drop() );
+  item->flags_.change( Core::OBJ_FLAGS::ATTACKABLE, is_attackable() );
```
frozenblit: shouldn't this be in a setter?

pol-core/pol/mobile/attack.cpp
```diff
@@ -9,21 +9,75 @@
...
-#include "pol/mobile/charactr.h"
 #include "pol/regions/guardrgn.h"
+#include "pol/ufunc.h"
 
 
 namespace Pol::Mobile
 {
+Attackable::Attackable( Core::UObject* obj ) : _opp( obj )
+{
+  if ( !_opp )
```
frozenblit: wouldn't the constructor be easier to understand if the _opp() is set explicitly at the end of the ifs, instead of reset to null?

pol-core/pol/mobile/charactr.cpp
```diff
@@ -3331,22 +3371,43 @@ void Character::attack( Character* opponent )
...
   {
     Core::gamestate.system_hooks.combat_advancement_hook->call(
         new Module::ECharacterRefObjImp( this ), new Module::EItemRefObjImp( weapon ),
-        new Module::ECharacterRefObjImp( opponent ) );
+        opponent.object()->make_ref() );
+  }
+
+  // special handling when attacking items
+  // they always hit
+  if ( auto* item = opponent.item() )
```
frozenblit, line 3453: did you consider splitting the attack method? seems a bit big, especially now with all the special casing of item vs mobile

pol-core/pol/mobile/npc.cpp
```diff
@@ -626,8 +626,9 @@ void NPC::readPropertiesForNewNPC( Clib::ConfigElem& elem )
   readCommonProperties( elem );
   readNewNpcAttributes( elem );
   readNpcProperties( elem );
-  calc_vital_stuff();
-  set_vitals_to_maximum();
+  // not yet entered the world, dont send vital pkts
+  calc_vital_stuff( true, true, false );
```
frozenblit: can you add a comment what is in each member? :D

pol-core/pol/multi/boat.cpp
```diff
@@ -372,9 +372,9 @@ void UBoat::send_display_boat( Network::Client* client )
     }
     if ( component == nullptr || component->orphan() )
       continue;
-    msg->Write<u8>( 0xF3u );
-    msg->WriteFlipped<u16>( 0x1u );
-    msg->Write<u8>( 0x0u );  // ItemData flag
+    msg->Write<u8>( 0xF3_u8 );
+    msg->WriteFlipped<u16>( 0x1_u16 );
+    msg->Write<u8>( component->is_attackable() ? 0x3_u8 : 0x0_u8 );  // ItemData flag
```
frozenblit: do we have an enum somewhere?

pol-core/pol/create.cpp
```diff
@@ -394,8 +394,9 @@ void ClientCreateChar( Network::Client* client, PKTIN_00* msg )
       chr->attribute( pAttr->attrid ).base( msg->SkillValue3 * 10 );
   }
 
-  chr->calc_vital_stuff();
-  chr->set_vitals_to_maximum();
+  // not yet entered the world, dont send vital pkts
+  chr->calc_vital_stuff( true, true, false );
```
frozenblit: comments would be nice

pol-core/pol/ufunc.cpp
```diff
@@ -629,8 +629,7 @@ void send_item( Client* client, const Item* item )
   if ( item->invisible() )
     flags |= ITEM_FLAG_HIDDEN;
 
-  auto pkt = SendWorldItem( item->serial, item->graphic, item->get_senditem_amount(), item->pos3d(),
-                            item->facing, item->color, flags );
+  auto pkt = SendWorldItem( item, flags );
   pkt.Send( client );
```
frozenblit, line 636: just a thought (not for now): we should send this to client.send(pkt), with pkt just providing a rendered byte buffer or so

pol-core/pol/item/item.cpp
```diff
@@ -1474,32 +1537,96 @@ void Pol::Items::Item::inform_enteredarea( Mobile::Character* whoenters )
...
+}
+
+void Item::destroy()
+{
+  if ( has_opponent_of() )
+  {
+    Mobile::Attackable self{ this };
+    for ( auto opp : *opponent_of() )
+    {
+      opp.remove_opponent_of( self );
```
frozenblit: does that clear the current opponent on the chr? does the item become an opponent_of of the chars?

pol-core/pol/mobile/charactr.cpp
```diff
@@ -3065,36 +3085,37 @@ void Character::set_opponent( Character* new_opponent, bool inform_old_opponent
...
+        opponent_.inform_engaged( this_att );
+      }
+      if ( mob )
+        mob->schedule_attack();
     }
-
-    send_highlight();
   }
+
+  send_highlight();
```
frozenblit: did you really want to send the highlight outside of the exit_signalled?

pol-core/pol/module/vitalmod.cpp
```diff
@@ -36,43 +38,51 @@ VitalExecutorModule::VitalExecutorModule( Bscript::Executor& exec )
...
-  }
-  return new BError( "Invalid parameter type" );
+  Mobile::Attackable att{ obj };
+  if ( !att )
+    return new BError( "Object is not damageable" );
+
+  bool send_dmg = send_damage_packet == 2 ? Core::settingsManager.combat_config.send_damage_packet
+                                          : ( send_damage_packet > 0 ? true : false );
+  if ( auto* chr = att.mobile() )
+    damage = chr->apply_damage( static_cast<unsigned short>( damage ), GetUOController(),
```
frozenblit: mobiles send the actual damage, while items send what was requested. Is that correct?

pol-core/pol/mobile/charactr.cpp
turleypol: Nope I think not :)

pol-core/pol/module/vitalmod.cpp
turleypol: Since no armor or anything else is used, the requested damage gets only capped by the hp. So requested==used. Mmmh but for consistency it's maybe better to just return it

pol-core/pol/mobile/attack.cpp
```diff
@@ -81,6 +81,15 @@ void Attackable::inform_engaged( const Attackable& engaged )
     item_->inform_engaged( engaged );
 }
 
+bool can_engage_item( const Character* chr, const Items::Item* item )
+{
+  if ( item->invisible() && !chr->can_seeinvisitems() )
+    return false;
+  if ( !chr->in_visual_range( item ) )
```
frozenblit, line 88: does this return true for worn items?

frozenblit (approved):

PR #896, opened 2026-07-13
Misc modernizations and improvments

turleypol:
Fixedallocator:
Type safe
General cleanup
Log facility:
Use move_only_function for queue
Message queue:
Added timeout wait method
TaskThreadPool:
Use move_only_function for queue 
DynTaskThreadPool:
Use move_only_function for queue
Added minimum number of workers
Instead of keep growing, the workers have now a timeout of 1min, remove these workers until minimum workers are reached
Curl commands:
Get rid of overhead due to DynTaskThreadPool changes
SQL service:
Use move_only_function for queue and get rid of overhead
Client transmit:
Use move_only_function for queue and get rid of overhead
Macos build:
Update to macos-latest

### DIFF
--- a/pol-core/clib/threadhelp.cpp
+++ b/pol-core/clib/threadhelp.cpp
@@ -357,18 +357,20 @@ void TaskThreadPool::init( unsigned int max_count, const std::string& name )
           {
             while ( !_done )
             {
-              _msg_queue.pop_wait( &f );
-              f();
+              try
+              {
+                _msg_queue.pop_wait( &f );
+                f();
+              }
+              catch ( std::exception& ex )
+              {
+                ERROR_PRINTLN( "Thread exception: {}", ex.what() );
+                Clib::force_backtrace( true );
+              }
             }
           }
           catch ( msg_queue::Canceled& )
           {
-          }
-          catch ( std::exception& ex )
-          {
-            ERROR_PRINTLN( "Thread exception: {}", ex.what() );
-            Clib::force_backtrace( true );
-            return;
           }
           // purge the queue empty
           std::list<msg> remaining;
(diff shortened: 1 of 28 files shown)


pol-core/clib/threadhelp.cpp
```diff
@@ -447,137 +451,159 @@ class DynTaskThreadPool::PoolWorker
...
       {
         ThreadRegister register_thread( _name );
+        ERROR_PRINTLN( "created pool worker {}", _name );
         auto f = msg();
         try
         {
           while ( !_parent->_done && !Clib::exit_signalled )
           {
-            _parent->_msg_queue.pop_wait( &f );
+            if ( !_parent->_msg_queue.pop_wait_for( &f, TIMEOUT ) )
```
frozenblit: just wondering: wouldn't it be better to have a slower ramp down? so if you have a peak, you keep threads up for slightly longer the greater the peak was. Right now after a peak you might remove many threads at once, which would require recreating them on the next peak

pol-core/clib/threadhelp.cpp
```diff
@@ -447,137 +451,159 @@ class DynTaskThreadPool::PoolWorker
...
+            {
+              std::lock_guard<std::mutex> guard( _parent->_pool_mutex );
+              // we timed out: can we retire?
+              if ( _parent->_live_threads > MIN_WORKER )
+              {
+                --_parent->_live_threads;
+                _retired.store( true, std::memory_order_relaxed );
+                ERROR_PRINTLN( "removed pool worker {}", _name );
```
frozenblit, line 530: ERROR_PRINTLN might flood the logs if there's a lot of deletion/recreation of threads

frozenblit (dismissed):

pol-core/clib/threadhelp.cpp
frozenblit: maybe a few things to try:
- add jitter i.e. TIMEOUT + random(5-20sec)
- limit how many workers per X seconds can be retired
- add a random chance of retiring per worker (0.75?) so not all workers retire at the same time after a peak

Might also wanna consider if we want to configure the timeout/min_workers for different queues

frozenblit (approved):

PR #898, opened 2026-07-19
Fixed shortcircuit optimization for consecutive jumps

turleypol:
Fixes #800 
The shared jump logic fails after the third ShortCircuit and jumps completely to the end.
Added minimal reproducer as test.

Fixed it by iterating over the bytecode rather then the ast to find consecutive jumps.

### DIFF
--- a/pol-core/bscript/compiler/optimizer/ShortCircuitCombiner.cpp
+++ b/pol-core/bscript/compiler/optimizer/ShortCircuitCombiner.cpp
@@ -1,10 +1,7 @@
 #include "ShortCircuitCombiner.h"
 
-#include "bscript/StoredToken.h"
 #include "bscript/compiler/Report.h"
 #include "bscript/compiler/ast/BinaryOperatorShortCircuit.h"
-#include "bscript/tokens.h"
-
 
 namespace Pol::Bscript::Compiler
 {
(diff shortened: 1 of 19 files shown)


frozenblit (dismissed):

pol-core/bscript/compiler/optimizer/ShortCircuitCombiner.cpp
```diff
@@ -46,10 +28,53 @@ void ShortCircuitCombiner::visit_children( Node& ) {}
...
   // dont visit recursivly just direct both sides
-  ShortCircuitCombinerChildVisitor lhs_child{ op, false };
+  ShortCircuitCombinerChildVisitor lhs_child{};
   op.lhs().accept( lhs_child );
-  ShortCircuitCombinerChildVisitor rhs_child{ op, true };
+  ShortCircuitCombinerChildVisitor rhs_child{};
   op.rhs().accept( rhs_child );
 }
 
+void ShortCircuitCombiner::optimize_jumps( CodeSection& code )
```
frozenblit: is this still a short circuit combiner optimization? because this is more like an optimizer stage, that just so happens to help the short circuit

frozenblit (dismissed):

frozenblit (approved):

PR #902, opened 2026-07-22
Fix uoconvert bug with UOPs and refactor code

frozenblit:


turleypol (dismissed):

pol-core/plib/clientfiles/tiledatacache.h
```diff
@@ -0,0 +1,99 @@
...
+
+namespace Pol::Plib
+{
+// One tiledata.mul entry's cached fields.
+struct TileData
+{
+  u8 height;
+  u8 layer;
+  u32 flags;
+};
```
turleypol, line 23: static assert the size, to make sure no padding is added which would then fail to load correctly

frozenblit: That's not used for loading files, it gets assigned from the files

turleypol: Ah ups yes

turleypol (approved):

PR #903, opened 2026-07-22
Fix ShortCircuit jump optimization corrupting assignments

murderinc:
The short-circuit jump optimizer threaded a logical jump past an opposite-type logical jump (offset++). That target pops the value on fall-through (Executor::ins_logical_jump), so skipping it left a stray value on the stack, which corrupts a following stack-addressed assignment:

    var x := ( a && b ) || ( c && d );

left x uninitialized whenever the first clause was false (i.e. the deciding clause is not the first operand). The same chain works as an if() condition or when declaration and assignment are split ( var x; x := ... ).

Only thread same-type logical jumps; keep targeting opposite-type ones so their fall-through pop still runs.

Adds testsuite/escript/opt/shortcircuit7 exercising the value / assignment context (companion to shortcircuit6's if()-condition tests).


pol-core/bscript/compiler/optimizer/CodeSectionOptimizer.cpp
frozenblit: does the RSV_JPMIFFALSE and RSV_JMPIFTRUE have the same issue?

turleypol:
Integrated into #904

PR #905, opened 2026-07-23
shortcircuit8: cover the && do..dowhile predicate (opposite-type JMPI…

murderinc:
…FTRUE)

shortcircuit8 exercised a ternary (JMPIFFALSE) and a || do..dowhile predicate, i.e. a same-type logical jump landing on the JMPIFTRUE. The remaining combination -- an && predicate, a JMPIFFALSE-type logical jump landing on the loop's JMPIFTRUE -- was untested. It corrupts the return value the same way (uninitialized object under the buggy peephole), so add r5/r6 driving a do..dowhile with an && predicate to close the last sub-case.


turleypol:
Integrated into #904

PR #907, opened 2026-07-25
Add configurable title separators (ServSpecOpt)

murderinc:
The core assembles a mobile's displayed title (paperdoll + AOS tooltip) as title_prefix + " " + name + " " + title_suffix, with a " " before "(race)" and "[guild]". Those separating spaces were hardcoded, so a shard whose title_prefix/title_suffix already carry their own separator (a trailing space, or a leading ", ") ends up with a doubled space or a " ,".

Add four ServSpecOpt string options - TitlePrefixSeparator, TitleSuffixSeparator, TitleRaceSeparator, TitleGuildSeparator - each defaulting to a single space, so existing output is byte-for-byte unchanged. Setting one empty lets the title strings supply their own separator (e.g. a suffix of ", Guildmaster" rendered directly against the name).

Docs (configfiles.xml) and corechanges.xml updated.


turleypol:
Can you add the new settings also to plib/baredistrofiles.cpp
(I think we really should in the future add these as extra example files to the repo)

docs/docs.polserver.com/pol100/configfiles.xml
```diff
@@ -361,6 +361,10 @@ Spell (int spellid)
 [InvulTag                       (0/1/2 {default 1})]
 [StartingGold                   (int {default 100})]
 [PrivacyPaperdoll               (0/1 {default 0})]
+[TitlePrefixSeparator           (string {default " "})]
```
turleypol, line 364: This file has at the top a last modified entry which is shown in the html

murderinc: Done.
Also added next PR with extended instructions for Claude, so it knows what to check next time.

turleypol (approved):

PR #915, opened 2026-07-28
Clamp AOS tooltip text before narrowing it to 16 bits

murderinc:
utf16.size() was cast to u16 before the range check, so a text longer than 65535 characters wrapped to a small value, passed the check unresized, and tripped the packet-size assertion in Write, which can shut the shard down. A script can set a name or description of any length.

The old 0xFFFF / 2 - 22 was not a mis-parenthesised ( 0xFFFF - 22 ) / 2 - that would overflow the packet, since the fixed overhead is 25 bytes, not 22.


turleypol:
Looking good but the conflicts need to be resolved

PR #951, opened 2026-08-23
Spread a world save across the save threads instead of one per file (Option A)

AsYlum-:
# Spread a world save across the save threads instead of one per file

A save got one thread per data file, so it took as long as its biggest file — and on a large
shard one file is most of the save. A measured example: `storage.txt` was 656MB of the 815MB
written, seven times the next biggest. However many threads such a shard had, the save waited
for storage.

The four files that are actually big are now cut into pieces that all share the task pool, so no
single file sets the pace, and the writing runs on a thread of its own. **Every file still comes
out byte for byte what a single thread wrote** — no format change, nothing to convert, external
tools that read the data files are unaffected.

## Numbers

End to end on a synthetic world shaped from the storage.txt above (189MB, 8 threads):

| | world stopped |
|---|---|
| one thread (`WorldSaveThreads=1`) | 545 ms |
| one thread per file — the old design's floor | 372 ms |
| this PR | **98 ms** |

The middle row is the point: it is `storage.txt` formatted single-threaded, and no amount of
per-file parallelism gets under it.

Microbenchmarks (`testing/testsave.cpp`): a CProp line 218 → 47 ns, an item's whole property
block 481 → 338 ns.

## The scheduler

New `pol-core/pol/saveparallel.{h,cpp}`. A `SavePart` is one file — or a pair written together,
like `pcs.txt` with `pcequip.txt` — as a piece count plus a `format( begin, end, writers )`
callback. `write_parallel` cuts every part into chunks, runs them all over the pool as one pool
of work, and appends the results in order.

- **Ordering.** Chunks are laid out part by part and written in that order, so each file receives
  its pieces exactly as a single thread would have produced them. Paired files are split in
  lockstep.
- **Bounded memory.** A sliding window caps chunks in flight at `2 × pool width`. Peak memory is
  window × chunk size, independent of world size — a 4GB shard holds no more than an 800MB one.
- **Writing.** A dedicated writer thread takes each finished chunk in turn, so producing text and
  handing it to the disk overlap instead of taking turns on one thread.
- Callers must not invoke it from a pool thread; it waits on pool work. `write_data` runs it on
  the async save thread, which is otherwise only waiting.

`storage.txt` is split at **root-item** boundaries, not per storage area: on the shard measured,
one area ("World Bank", 5630 player bank boxes holding 2.4M items) was 99.7% of the file, so
per-area splitting would have left one thread with all of it. 7231 root items, largest 0.67% of
the file — it balances well.

## Less work per object

- `PropertyList::printProperties` no longer builds a temporary `std::string` per CProp just to
  hand it to `StreamWriter::add`. Same for character attributes and skills, custom house tiles,
  spellbook bits and map pins. The shard measured writes 6.9M CProp lines per save.
- `ObjectHash::for_each_character` replaces three full walks of the object hash. That hash holds
  every item on the shard, so each pass walked millions of nodes to reach a few thousand mobiles.
  Character and item serials are disjoint ranges, so this visits only that stretch of the map. It
  starts at `begin()` rather than `lower_bound( CHARACTERSERIAL_START )` deliberately, making it
  exactly the old walk minus the items it used to test and discard.
- `StreamWriter` hands blocks larger than its buffer straight to the file instead of copying them
  in first, and flushes at 1MB rather than 32KB. Inline capacity stays 32KB — thirteen writers
  live together in `SaveContext`.
- The tail flush and the `.ndt` → `.txt` commit now fall outside the part the world is stopped
  for, and are timed separately.

## Added

`pol.cfg LogWorldSaveDetails` (default 0) logs where a save went:

```
Worldsave ok: 420 ms world stopped (312 ms of it writing), 2 ms flushing, 61 ms committing
  parts (work, summed over the threads that shared it):
    storage      2284 ms
    items         310 ms
    ...
  files:
    storage      688301056 bytes
    ...
```

`pol.cfg WorldSaveThreads` (default 0) sets the pool width. The default now comes from
`Clib::available_cpus()`, which accounts for the cgroup cpu quota and the affinity mask on Linux
— `std::thread::hardware_concurrency()` ignores both, so in a container POL sized the pool from
the *host's* cpu count and started more save threads than the container was allowed to run. The
startup line prints all three numbers.

`uo::SaveWorldState()` gains `ElapsedMillisecondsWriting`, `ElapsedMillisecondsFlush`,
`ElapsedMillisecondsCommit`, and `Files` / `Tasks` structs. Additive only. Note `Tasks` is
**work summed across threads**, not wall time, so its entries deliberately exceed
`ElapsedMilliseconds`.

## Worth a close look

- **`write_parallel`'s threading.** Reviewed hardest by `save_parallel_test`, which holds the
  parallel output against a single-threaded run over ordinary, paired and empty parts.
  One race was found and fixed during development: the writer could reach a chunk the launcher
  had not submitted yet, and `get()` on a spent future throws — a spurious save failure, rare
  enough that 25 stress runs missed it. Gated on an explicit launched count now.
- **The writer thread earned nothing measurable on my machine** (95–103 ms vs 92–99 ms before,
  and no change across window depths 2/4/8). Both stages sit at ~75% duty there, which points at
  memory bandwidth rather than scheduling. It should help where format and write overlap worse —
  the shard measured shows only 45% overlap against my 65% — but that is unverified. Reasonable
  to ask for it to be dropped if it does not show up on other hardware.
- **`Character::printForSave`** replaces the two `friend void Core::write_characters/write_npcs`
  declarations. A lambda is not covered by the enclosing function's friendship, and a public
  "one save unit" method reads better than widening the friendship.
- **Linux is unbuilt.** `-fPIC` → `-fPIE` for POL's own targets (nothing here is a shared
  library; `-fPIC` on an executable costs a GOT indirection per global and general-dynamic TLS
  per `thread_local`, which the save path is full of). Also a new opt-in `DISABLE_HARDENING`
  **for benchmarking only** — it exists to put a number on what `-fstack-protector` and
  `_FORTIFY_SOURCE` cost, since Windows has always built with `/GS-`. Needs CI to confirm both.
- `BENCHMARK_ENABLE_WERROR OFF` in `cmake/Benchmark.cmake`: `-DENABLE_BENCHMARK=ON` did not build
  at all on current MSVC before this.

## Verification

- **Byte identity against master**, which is the gate for all of this: built both binaries, ran
  the same world through load-and-resave (`POLCORE_TEST_RUN=3`) with each, compared all 14 data
  files. Identical, `pol.txt` aside, where the only lines that differ are `CompileDateTime` and
  the game clock.
- 91/91 ctest, `shard_test_roundtrip` included. 40 repeat runs of the threaded save tests, clean.
- New: `save_parallel_test` (scheduler ordering), `storage_print_test` (storage.txt parallel vs
  single-threaded), and assertions on the new `SaveWorldState()` members in
  `testpkgs/misc/test_polcore.src`.
- No thread sanitizer — MSVC has none. `-DENABLE_TSAN=ON` on Linux would be worth one run.

## Also fixed in this PR

Not introduced by this PR - SaveContext::ready() dates to #732 and the callback's lock to #745 - but this PR makes it far easier to hit, so it's fixed here.

The cycle. The scripts thread holds the world lock for an entire step_scripts() pass, so a script calling uo::SaveWorldState() reaches write_data with that lock held. write_data opens with a blocking SaveContext::ready(), waiting on the previous save's finished future. That future only became ready after the previous save's completion callback had run - and the callback's first line is Core::PolLock lck, because it revives the script that asked for the save.

```
scripts thread                      save thread
──────────────                      ───────────
pol.cpp:573    PolLock ────────────► holds polsem
pol.cpp:579    step_scripts()
uomod.cpp:2916 SaveWorldState()
savedata.cpp:431 ready()
savedata.cpp:188   finished.wait() ═╗
                                    ║  savedata.cpp:624  callback( stats )
                                    ║  uomod.cpp:2920    Core::PolLock ← blocked
      ▲                             ╚═════════════════════════╪
      └───────────────────────────────────────────────────────┘
```

The save thread waits for a lock held by the thread waiting for the save thread. Two scripts saving in the same pass is enough. Shutdown hangs the same way - pol.cpp:1238 takes PolLock, then calls write_data.

Why this PR widened it. The critical section now ends much earlier, so the scripts thread is released well before the tail (flush, 13 commits, callback) finishes. The window where a second save lands inside a running save's tail got much bigger.

The fix. finished becomes ready once the files are committed, and the completion callback is handed to a dedicated single-thread worker instead of running inline before it. The future the scripts thread waits on no longer spans a lock acquisition, so the cycle can't form. Nothing waits on that worker, so it's free to block for the lock as long as it needs.

Deliberately not the task thread pool: the scripts thread holds the world lock while waiting on that pool for the next save's objects, so a callback parked there would occupy the thread that wait depends on - and with WorldSaveThreads=1 it's the only one. That would relocate the deadlock, not remove it.

The rest of the save path takes no lock anywhere, so the wait for the critical part was already safe, and the parallel scheduling is unchanged.


pol-core/clib/streamsaver.cpp
```diff
@@ -31,17 +33,29 @@ StreamWriter::~StreamWriter() noexcept( false )
   }
 }
 
+void StreamWriter::write_block( const std::string_view& text )
```
turleypol: Pass by value when using string_view

pol-core/clib/streamsaver.h
```diff
@@ -57,21 +101,62 @@ class StreamWriter
...
+    if ( _file && _mbuff.size() > FLUSH_THRESHOLD )
+      flush();
+  }
+  void flush();
+  /// Write one block straight to the file. Callers are responsible for the buffer being empty.
+  void write_block( const std::string_view& text );
+
+  /// Inline capacity, which every live writer costs wherever it is allocated -- and a world save
+  /// holds thirteen of them at once. Kept small on purpose; the buffer grows on the heap.
+  static constexpr size_t INLINE_SIZE = 0x8000;
```
turleypol: Mmmh that's the start size of the buffer, but every spot flushes based on the threshold. Thus each writer will in the end be at least the size of threshold. Why not directly reserve the threshold?

pol-core/pol/mobile/charactr.cpp
```diff
@@ -716,7 +716,7 @@ void Character::printProperties( Clib::StreamWriter& sw ) const
     sw.add( "PartyCanLoot", party_can_loot() );
   for ( const auto& rt : reportable_ )
   {
-    sw.add( "Reportable", fmt::format( FMT_COMPILE( "{:#x} {}" ), rt.serial, rt.polclock ) );
+    sw.add( "Reportable", Clib::hexintv( rt.serial ), rt.polclock );
```
turleypol: Use .raw? The stringview variant still needs a buffer to be able to return a view

pol-core/pol/savedata.cpp
```diff
@@ -225,75 +238,100 @@ void WriteGottenItem( Items::Item* item, Clib::StreamWriter& sw )
...
+          items.push_back( item );
       }
     }
   }
+  return items;
+}
+
+SavePart items_part( const std::vector<Items::Item*>& items, Clib::StreamWriter& sw )
+{
+  return { "items",
```
turleypol: A bit hard to read. With designated initializer it would be better

pol-core/pol/saveparallel.cpp
```diff
@@ -0,0 +1,257 @@
...
+
+  size_t next = 0;
+  while ( next < window )
+  {
+    launch( next, next );
+    ++next;
+  }
+  pipe.launched = next;  // nothing else is running yet
+
+  std::thread writer;
```
turleypol: jthread?

pol-core/pol/saveparallel.cpp
```diff
@@ -0,0 +1,257 @@
...
+  try
+  {
+    writer = std::thread( write_finished_chunks );
+    // Slot `next % window` last held chunk `next - window`, so it is free to reuse as soon as the
+    // writer has dealt with that one. Waiting on that is what bounds the memory a save holds:
+    // `window` chunks, whatever the size of the world.
+    while ( next < chunks.size() )
+    {
```
turleypol, line 223: Why can't we just push all the chunk into the queue at once?

pol-core/clib/streamsaver.cpp
AsYlum-: Done — every `const std::string_view&` in `StreamWriter` is now `std::string_view`:
`write_block`, both `add` overloads, `key`, `comment` and `append`.

pol-core/clib/streamsaver.h
AsYlum-: You're right about where the buffer ends up, but `INLINE_SIZE` was not a reserve — it was the
template argument of `fmt::basic_memory_buffer<char, N>`, i.e. `N` bytes stored *inside the
object*. `SaveContext sc;` is a stack local (`savedata.cpp`) holding thirteen writers by value,
so at `0x8000` that was already ~416KB of the save thread's stack, and at `FLUSH_THRESHOLD` it
would have been 13MB — an overflow on Windows' 1MB default thread stack.

So I did what you're after, but on the heap: dropped the inline capacity entirely (fmt's default
is 500 bytes) and `_mbuff.reserve( FLUSH_THRESHOLD )` in the file-backed constructor. The end
state is now taken in one allocation instead of growing into it through ~9 of fmt's 1.5×
reallocations, and `SaveContext` on the stack drops from ~416KB to ~1KB.

Detached writers deliberately get no reserve: they never flush, they grow to whatever one chunk
needs and are then reused via `reset_buffer()`, which keeps the capacity.

Cost is ~13MB of transient heap per save — the nine small files now hold a 1MB buffer they never
fill. Cheap next to the world they're saving, but worth naming.

pol-core/pol/mobile/charactr.cpp
AsYlum-: Correct, and it was worth more than the one line: `hexintv` formats into a `thread_local
fmt::memory_buffer` and returns a view of it, so every `add( key, hexintv( x ) )` formatted
twice. There were 15 of them, all in the save path, and `uobject.cpp` (Serial / ObjType /
Graphic / Color) runs for every object in the world.

Rather than three lines of `key()`/`raw()`/`eol()` per site, I wrapped exactly that in
`StreamWriter::add_fmt`:

```cpp
template <typename S, typename... Args>
void add_fmt( std::string_view k, S&& formatstr, Args&&... args )
{
  key( k );
  raw( std::forward<S>( formatstr ), std::forward<Args>( args )... );
  eol();
}
```

All 15 sites are now `sw.add_fmt( "Serial", FMT_COMPILE( "{:#x}" ), serial )`, formatting
straight into the writer's buffer, and `Clib::hexintv` is gone from `strutil.h` — nothing else
used it. Byte-identical output: the format string is the one `hexintv` used, and `add`/`add_fmt`
frame it identically.

The three-argument `add()` stays — `proplist.cpp`'s CProp line is its remaining user, which is
what it was added for.

pol-core/pol/savedata.cpp
AsYlum-: Done, in all three `SavePart` factories: `items_part`, `mobiles_part` and `Storage::save_part`.

pol-core/pol/saveparallel.cpp
AsYlum-: Switched. Both `join()` calls stay explicit, though, and I left a comment saying why: `drain()`
may only run once the writer has stopped, since both call `get()` on the same futures, and
`writer` is declared outside the `try` so its destructor would fire *after* `drain()`. The
`jthread` is a backstop for a path that doesn't currently exist rather than the mechanism.

pol-core/pol/saveparallel.cpp
AsYlum-: This will be long answer but I think it should clarify what's happening :) 

### What the window actually is

Three things are in play, and only the third is a queue:

1. **The pool queue** — unbounded, `checked_push` → `_msg_queue.push_move`.
2. **The slots** — `window` sets of detached `StreamWriter`s, allocated once at
   `saveparallel.cpp:104-107` and recycled.
3. **The writer thread** — walks `head = 0..chunks.size()` in order, and for each one takes
   `slots[head % window]`.

That modulo is the whole mechanism. Slot `i` is reused by chunks `i`, `i + window`,
`i + 2·window`… so before chunk `n` can be launched, chunk `n - window` must already have been
drained out of that slot. Hence the wait at line 223:

```cpp
pipe.progress.wait( lock, [&]() { return pipe.written + window > next || pipe.failed; } );
```

which is exactly "chunk `next - window` has been written". The number of chunks holding
formatted text is therefore never more than `window`, regardless of how big the world is.

### Why it can't just be removed

The two halves run at structurally different speeds, and that's deliberate:

- **Formatting** is N-way parallel — that's the point of the commit.
- **Writing** is serial and can't be otherwise: a file must receive its chunks in order, so one
  thread appends them one after another.

A producer that is N× faster than its consumer, feeding an unbounded queue, has one outcome —
the queue grows until the input is exhausted. Here the "queue" is formatted save text, so the
backlog converges on *the entire save*.

Rough numbers for a 2GB save on 8 threads: `cut_into_chunks` yields up to
`width × CHUNKS_PER_THREAD` per part × 4 parts = 512 chunks, so ~4MB each.

| | peak formatted text held |
|---|---|
| today, `window = 16` | ~64MB |
| all chunks pushed | ~2GB |

And that 2GB is added *while the world is stopped*, on top of the world itself — the worst
possible moment to double the process's footprint.

### The second cost: the buffers stop being reusable

`window` slot buffer sets are allocated once. `reset_buffer()` clears them but keeps capacity,
so from the second round onward chunk formatting does **zero** buffer allocation.

One task per chunk means one buffer set per chunk — 512 of them, each allocated cold and grown
into by fmt. You'd lose the reuse entirely and pay allocation churn per chunk, on the threads
whose speed the whole design depends on.

### Why `threads × 2` specifically

`WINDOW_PER_THREAD = 2` is double buffering: every thread can have one chunk formatting and one
finished-and-waiting. That's the minimum depth at which no pool thread ever idles waiting for a
slot.

So a deeper queue can't buy throughput. If the writer keeps up, the extra depth is never used;
if it doesn't, the disk is the bottleneck and queueing more work in front of it just converts
RAM into nothing. The knob is there (`WINDOW_PER_THREAD`) if a measurement ever says otherwise.

### The bounded variant is worse

The obvious middle ground — push all 512 tasks, have each block until a slot frees — moves the
wait *into* the pool. Now pool threads sit on a condition variable for the length of the save,
and `gamestate.task_thread_pool` is shared with the rest of the server. You'd trade a bounded
memory footprint for occupied threads and get nothing back.

PR #957, opened 2026-09-02
Load npcdesc.cfg and itemdesc.cfg recursively or as suffix

Kukkimonsuta:
This PR adds ability to load additional npcdesc/itemdesc config files which should help reducing clutter.

Files that should be loaded (order is undefined atm):
- `config/npcdesc.cfg` (current)
- `config/**/*.npcdesc.cfg`
- `<pkg>/npcdesc.cfg` (current)
- `<pkg>/**/*.npcdesc.cfg`

Notes:
- `bool FindNpcTemplate( const char* template_name, Clib::ConfigFile& cf, Clib::ConfigElem& elem )` Doesn't seem to be used anywhere anymore, so instead of porting it i removed it.
- `pol-core/pol/module/polsystemmod.cpp` member `npcdesc` - not sure what to do here
- it's bit weird that these are getting iterated in multiple places, but I'm not brave enough to touch on that

### DIFF
--- a/pol-core/clib/fileutil.cpp
+++ b/pol-core/clib/fileutil.cpp
@@ -22,17 +22,22 @@
 
 namespace Pol::Clib
 {
-std::string normalized_dir_form( const std::string& istr )
+std::string normalized_path_form( const std::string& istr )
 {
   std::string str = istr;
 
+  std::string::size_type bslashpos;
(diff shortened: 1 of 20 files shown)


Kukkimonsuta:
Abandoning - this version doesn't support recursive packages which somewhat alleviate the issue this is trying to solve.

PR #962, opened 2026-09-05
Repaint the client's weather and light after anything that resets them

murderinc:
The client stops drawing weather when it is sent a new position, and the season packet resets its light level. Both go out from many places, and only move_character_to handled the consequence - it nulls the client's remembered weather region so its own tellmove repaints. Everywhere else the sky stayed wrong until the character crossed into another weather region, teleported, or the weather changed on its own.

Both halves are measured, not inherited. The tree only ever hedged on the weather half - "send_goxyz seems to stop the weather" beside the null in move_character_to, and dave's 2003 note in check_weather_region_change, "send_goxyz causes weather effects to stop if character is walking/running too". Sending a bare, well-formed 0x20 to a standing character in the rain, with no repaint behind it, stops the weather: the sky goes clear and stays clear. So the effect is real and the walking/running qualifier is not limiting. For the light, send_season_info's own comment says the season packet resets it, and an undated POL097 entry records the intent that the core "sends light level whenever season info is sent to the client".

Repaints added at: the client's resync request, death, resurrection, a graphic, colour or facing change, the gargoyle fly toggle, the client version handshake, and a boat crossing realms. Login already had one, but it went out before the season packet and the final send_goxyz and so undid itself - that one moves rather than appears. Only the handshake rests on the season premise; it sends no position packet at all. The rest stand on send_goxyz alone.

The boat repaint has to sit below the branch that sends goxyz to pre-7.0.9.0 clients, or that undoes it, and has to be forced: region grids are zero-filled per realm and the lookup indexes one shared vector, so an unpainted zone in either realm is the same background region and an unforced check sees no change.

send_season_info knew the season packet resets the light but only restored it for a weather region carrying a LightOverride - and where it declined, a character holding its own override, the light stayed reset. It now re-sends gd->lightlevel, which is what the client was last told in every case.

That exposed the bug it had been masking: a personal override expiring inside such a region never restored the region's level, because both places that decide who owns the light skipped the character on the assumption the client already had it. Those two are now weather_region_owns_light(). regen_stats open-codes the expiry half as well and is left alone; only the skip was duplicated.

update_weatherregion now also accepts a client with no remembered weather region. It gated on the pointer matching the region being changed, so the two calls below
- which null it - would have made that client invisible to SetRegionWeatherLevel, the main way weather is driven. Inferred, not stated: the commented-out 2003 line in that same function looks like the same edge from the other side.

setseason and SendOverallSeason keep their documented contract - resending the light and weather is the script's job - but no longer leave the core recording state the client has lost. Both are invalidated instead: the weather region is nulled, and the light level set to -1, which VALID_LIGHTLEVEL rejects so it can never be sent by accident. No packet goes out, so the contract holds. The repair is not immediate, though: it lands on the character's next region change or the next SetRegionWeatherLevel.

ClientGameData::lightlevel now starts at -1 rather than 0, and clear() resets it. Zero is a real level, so the old value claimed the client had been told something it never had - a character logging into a region whose level is genuinely 0 was never sent one at all.

chr.facing :=, chr.setfacing() and npc::Face() now call on_facing_changed only when the facing actually changed. face() reports success either way, so a script turning to face a target it already faced was resending the position to itself and its movement to onlookers every call. Scripts using a same-direction turn as a resync nudge lose that side effect.

The repaint is not free where facing changes are frequent: a client announces the weather on every packet it accepts, so each repaint is also a journal line while weather is falling. A clear sky is silent - that packet carries no message.

Deliberately unchanged: a boat step or turn sends goxyz to every traveller on a pre-7.0.9.0 client, so those clients still lose the weather while sailing even within one realm. The repaint belongs in do_tellmoves, the single end-of-move point for move_to, move and turn, and wants testing against a real client first.

Claude-Session: https://claude.ai/code/session_01XWxMZWAVo2agN6axNF22o2

### DIFF
--- a/docs/docs.polserver.com/pol100/configfiles.xml
+++ b/docs/docs.polserver.com/pol100/configfiles.xml
@@ -13,7 +13,7 @@
       <list>elipsis ... : repeat entries are allowed.</list>
       <bottom>This means the punctuation in these cases should NOT be in the actual config files used by POL, they only appear here for information purposes. Exception: curly braces { } are used to define an element in a config file. These must be present in the actual file.</bottom>
     </desc>
-    <datemodified>09/05/2026</datemodified>
+    <datemodified>09/06/2026</datemodified>
 </fileheader>
 
 
@@ -443,8 +443,9 @@ Spell (int spellid)
   <explain><i>WhisperRange:</i> definies the range of whispered speech</explain>
   <explain><i>YellRange:</i> definies the range of yelled speech</explain>
   <explain><i>CoreSendsSeason:</i> Determines if the core should send season packet on char creation/logon/reconnect,
-        realm change and the client version handshake, based on the season entry in realm.cfg. The season packet resets
-        the client's light level, so the core restores it behind each packet it sends.</explain>
+        realm change, the client version handshake and resurrection, based on the season entry in realm.cfg. The season
+        packet resets the client's light level, so the core restores it behind each packet it sends. The one on
+        resurrection is there because a season sent while the character was dead does not stick.</explain>
   <explain><i>CoreHandledTags:</i> bitfield to determine which tags are displayed on singleclick, current used bits are:<br/>
     0x1  [title_guild]<br/>
     0x2  [frozen]<br/>
--- a/docs/docs.polserver.com/pol100/configfiles.xml
+++ b/docs/docs.polserver.com/pol100/configfiles.xml
@@ -13,7 +13,7 @@
       <list>elipsis ... : repeat entries are allowed.</list>
       <bottom>This means the punctuation in these cases should NOT be in the actual config files used by POL, they only appear here for information purposes. Exception: curly braces { } are used to define an element in a config file. These must be present in the actual file.</bottom>
     </desc>
-    <datemodified>08/15/2026</datemodified>
+    <datemodified>09/05/2026</datemodified>
 </fileheader>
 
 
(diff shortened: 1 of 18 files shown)


turleypol:
Could you please let it create a few tests for it?

murderinc:
10 of the 16 changed behaviours have a case in `testsuite/pol/testpkgs/client/test_client_weather_repaint.src`. 

The rest:

  **Login and the version handshake** — need test-client work first, not just a
  case: `startclient` selects the character before attaching a brain, so nothing
  sent during login can raise an event. Needs the client to record what it last
  received and a todo to ask it afterwards. That would cover the two light fixes
  below too, so it is one piece of work rather than three.

  **Gargoyle flight toggle** — a `0xBF` subcommand via the `raw_packet` todo, as
  `weather_repaint_resync` does. Needs the subcommand id and a gargoyle body.

  **Boat crossing realms** — `test_boat.src` already moves a boat between realms,
  `test_client_boat.src` already sails one with a client aboard; this is the two
  combined.

  **Two light fixes** — the third is covered by `light_restored_after_season`.
  An override expiring inside a weather region carrying a `LightOverride` is
  blocked on data: `TestWeatherStorm` is the only weather region and carries
  `-1`, the case the fix is not about. Logging into a zero-level light region
  needs the login machinery above.

  **`npc::Face()`** — same guard as `facing_unchanged_is_quiet`, for an NPC.
  Needs a test NPC calling `Face()` and a second client asserting silence.

turleypol (approved):

PR #965, opened 2026-09-13
Read a world faster, and report where the time went

AsYlum-:
Startup on the measured shard — three million objects across 800 MB of data files — spent **65 seconds** in `read_data()`. It is **35** now, and the gain is near enough uniform across the data files, which is what says it came from the per-element work every file shares rather than from any one file's handling.

## The problem

Nearly all of `read_data()` went on work every element pays regardless of what it turns out to be. Two containers dominated:

| | cost per element | why |
|---|---|---|
| `ConfigElem` properties | a tree node + up to two heap strings **per property line**, built and freed for every element | `std::multimap`, rebuilt millions of times |
| `readProperties()` lookups | ~90 fixed name probes against the ~12 an element carries | most probes paid a full tree descent in order to miss |

### 1. `22c84f7` Read a world without rebuilding a property tree for every element

`ConfigElem`'s properties become a flat store: a run of hashes scanned in order, with the name compare only confirming a hit. `clear()` forgets the live count and removal tombstones the slot, so the vectors and their string buffers are reused rather than reallocated — the loader stops allocating for properties at all.

`PropKey` carries a name with its hash and converts implicitly from what call sites already pass, so literal names are hashed at compile time and **none of the ~900 call sites changed**.

Smaller wins on the same path:

- the line reader uses `memchr` over its own buffer instead of a locking `fgets` plus a copy, and   reads binary rather than letting the CRT translate CRLF
- a line is validated as utf8 only once a high bit shows up, which for world data is almost never
- `items.txt` and `storage.txt` share one stack of the containers the loader is inside, pushed for every container rather than only top-level ones; `storage.txt` had none at all, so all 2.4 M of its elements paid an object hash descent
- `ObjectHash::Insert` makes one tree descent rather than two
- the zone integrity scans in `add_item_to_world` and `SetCharacterWorldPosition` are `paranoid` rather than plain asserts — `INC_PASSERT` is on in release builds and they scan a 64×64 zone linearly, so filling a dense one cost O(items²)

### 2. `2b7023d` Report where reading a world spends its time

New `pol.cfg LogWorldLoadDetails` (default 0) breaks each data file into parsing, creating the objects, reading their properties and placing them — with the object hash split out of creation and custom properties out of `readProperties` — as a total and per element. Off, it reads no clock at all.

It paid for itself immediately. Placing an equipped item cost **27us** against a ground item's **0.28us**: `check_test_scripts` built a `ScriptDef` for `equiptest.ecl` in every package and asked `script_loaded()` about each one, per equipped item, so a shard with fifty packages paid fifty lookups to be told "no" forty-nine times. Locating them once is worth **2.8 s** of the 42.

It also ruled two things out, which is the other half of what a report is for: object hash insertion is 0.26us per object, and value interning already finds an existing entry 93% of the time. Neither was worth changing.

### 3. `db0c853` Keep an object's custom properties in a sorted vector, not a tree

A property list is a member of every `UObject`, and the measured shard carries eight and a half million custom properties — that many red-black-tree nodes whose pointers and colour outweigh the two handles they hold. The server's own report puts its objects at **1065 MB where they were 1396 MB**, and the process is smaller by more again: the allocator's per-allocation header and the fragmentation left by that many small allocations are real but not something `memsize` models.

**Order is the thing that had to be preserved, not the container.** A save writes properties in whatever order this iterates, and a vector sorted by the comparison `std::map` was given iterates the same way. `shard_test_roundtrip` is what holds that.

### 4. `79f4c83` Keep the case-insensitive string helpers together

Review feedback on the first commit. `ci_hash` and `ci_equal` move to `clib/maputil.h`, beside `ci_cmp_pred` — the same comparison in the form an ordered container takes — and the converter that turns any of the string shapes into a `string_view` comes out of `ci_cmp_pred` so all three share it. `ci_hash_pred` and `ci_eq_pred` join them: the pair an unordered container needs where an ordered one takes `ci_cmp_pred` alone. `std::hash<PropKey>` hands out the hash the key already carries.

The hash itself stays hand-written, because `std::hash` is neither case-insensitive nor constexpr and both matter here: a property name is matched regardless of case, and the ninety literal names `readProperties()` probes fold to their hashes at compile time, leaving the scan comparing numbers.

### 5. `10dcfd5` Decide a property probe, and a comment line, before doing the work

Two small things on the same path, both measured on the shard above.

`ConfigProps` keeps a 64-bit mask of the names an element holds, one bit per hash, and `find()` answers from it before scanning. An element carries about a dozen properties and is probed with some ninety names, so most probes now cost two instructions instead of walking every slot. Across the load `readProperties` went **11004 ms to 9972 ms**, and down in all seven files that report it — storage.txt 7629 to 6851, npcs.txt 1518 to 1425, items.txt 996 to 893.

The config reader also decides that a line is blank or a comment **before** `splitnamevalue()` takes it apart, rather than splitting it into a name and value and then dropping both. The trimmed view is what goes to the split, so a real line is not scanned for leading whitespace twice; `ltrim_view()` is inline in the header because nothing links with interprocedural optimization. Parsing across the load: **5743 ms to 5608 ms**.

What the same measurements ruled out, which is the other half of what a report is for:

- The per-item duplicate-serial probe in `read_item()` looked redundant, since `ObjectHash::Insert` detects a duplicate on the descent it has to make anyway. Removing it moved 1.6 s out of the unaccounted column and put 1.4 s **into** the object hash: the probe was warming the cache for the insert that follows it on the same key. Worth 0.2 s net, so it stays as it is.
- Cutting the per-item `find_itemdesc()` calls from four to two was worth 30 ms. `desctable` is small enough to stay in cache, so a lookup there is nothing like the 2.4 M-node object hash.
- Single runs on a live shard vary by 400-600 ms on the total and about 200 ms per phase, which is the floor anything smaller has to clear before it can be called a gain.

## Compatibility

- Nothing about the data files changes and nothing about what is loaded from them: the same objects come back with the same properties, in the same order.
- `PropertyList::properties` is `protected` and nothing subclasses `PropertyList`, so the container change is confined to `proplist.cpp`. One behavioural note for future work: `setprop()` on a new name now invalidates references into the list, where the map did not. Nothing holds one — `getprop` copies out.


pol-core/clib/cfgelem.h
```diff
@@ -7,12 +7,164 @@
...
+#include <string_view>
+#include <utility>
+#include <vector>
 
 
 namespace Pol::Clib
 {
+/// Case-insensitive FNV-1a over ASCII. constexpr so a literal name at the call site is hashed
+/// at compile time. Never returns 0: ConfigProps spends that value on marking a removed slot.
+constexpr u32 ci_hash( std::string_view s )
```
turleypol: std::hash ?

pol-core/clib/cfgfile.cpp
```diff
@@ -44,9 +44,19 @@ size_t ConfigElemBase::estimateSize() const
   return type_.capacity() + rest_.capacity() + sizeof( _source );
 }
 
+size_t ConfigProps::estimateSize() const
+{
+  size_t size = Clib::memsize( hashes_ ) + Clib::memsize( by_name_ ) +
+                entries_.capacity() * sizeof( entries_[0] );
+  // Tombstoned slots past the live count still hold their string buffers, so they count too.
+  for ( const auto& entry : entries_ )
```
turleypol: There is I think a memsize overload which takes a vector and a lambda for these purposes

AsYlum-: switched to `Clib::memsize( entries_, lambda )`

pol-core/clib/cfgelem.h
AsYlum-: Can't use std::hash directly — it hashes raw bytes, so Serial and serial would miss each other, and it isn't constexpr in any stdlib, so the ~90 literal property names readProperties() probes couldn't fold their hashes at compile time (that folding is the point: the runtime loop then only compares u32s). ConfigProps also reserves hash 0 as its tombstone marker. 

Moved ci_hash/ci_equal into clib/maputil.h next to ci_cmp_pred so the case-insensitive string helpers live together, added ci_hash_pred/ci_eq_pred there for unordered containers, and exposed the key's hash under the standard name via std::hash<PropKey>.

turleypol (dismissed):

turleypol (approved):

PR #966, opened 2026-09-14
Load boat components into the slots of their shape

murderinc:
`move_components()` and `transform_components()` pair `Components` with the component lines of the boat's shape in boats.cfg by index, and nothing checks that the two line up. A boat whose saved list is one entry out sends every component after that entry to another component's offset, on every step and turn. When the list is one too long, the last component has no shape line left to pair with and stays where it is.

`UBoat::align_components()`, called from `readProperties()` once both lists are read, rebuilds `Components` against the shape.

**How a save gets there**

`readProperties()` reads the `Traveller` lines first, and puts an item whose objtype `BoatShape::objtype_is_component()` accepts into `Components`, ahead of the `Component` lines that follow. Before #919 anything that moved a component as an ordinary item, such as a script `MoveObjectToLocation()` or a staff member dragging it, also registered it in `travellers_`. It was then saved under both keys, and the next load listed it twice with the extra copy first. Every later save writes that duplicate back out. #919 made `register_object()` refuse a component, but nothing repairs a save that already has one. A save can also come up short: `uo::DestroyItem()` does not refuse a component, and a destroyed one is not saved.

A small boat saved as

```
Component 0x5aca0a45   tillerman
Component 0x5aca0a45   tillerman, again
Component 0x5aca0a46   port plank
Component 0x5aca0a47   starboard plank
Component 0x5aca0a48   hold
```

had its tillerman at the port plank's offset, the port plank at the starboard plank's, the starboard plank at the hold's, and the hold left wherever the boat was when the list went wrong.

**The fix**

- Duplicates collapse to their last entry, which is the component's own `Component` line. Keeping the first would put a duplicated rope or hull piece in the first slot of its objtype and push the others down one.
- Each slot then takes the first remaining component of its objtype, so components sharing an objtype keep their relative order and each keeps a slot.
- A missing component leaves a null `ItemRef` in its slot. Every loop over `Components` already skips one, and `printProperties()` does not save it.
- A component with no slot is kept at the end, as before. It is saved, and it does not move.
- A boat whose `multiid_` has no boats.cfg shape is left alone, since `boatshape()` would assert.

**Behaviour change**

- A misaligned boat is put right as the world loads. A well-formed save loads into the list it had.
- When the missing component shared its objtype with later ones, such as one of several ropes, those each move up one slot and the last of them stays empty. The save does not record which slot was lost.

Scope is load only. `create_components()` can also come up short of the shape when a component fails to create or place. That is the same pairing problem, and is not touched here.

**Testing**

`boat_load_test` in `pol-core/pol/testing/testboat.cpp`, run by `unittest_pol`. It loads a boat from a hand-written `Multi` element, the way the world load reads one, then moves or turns it. It checks each component's offset, the graphic `transform_components()` gives it, and how many `Component` lines `printProperties()` writes back.

18 checks over six saves: a component listed twice, the same save turned, one listed as a traveller as well, two components sharing an objtype with one of them listed twice, a missing component, and a well-formed save. Without `align_components()`, 16 of the 18 fail.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


turleypol (commented):
I guess you have to message Asylum to stop a moment :)

turleypol (dismissed):

pol-core/pol/multi/boat.cpp
```diff
@@ -1333,6 +1334,44 @@ void UBoat::rescan_components()
   }
 }
 
+/**
+ * Puts the components read from a save into the slots of the boat's shape, which move_components()
+ * and transform_components() pair them with by index. A component listed more than once keeps its
+ * last place: an extra copy is first read from the travellers, ahead of the components, and every
+ * later save keeps it ahead.
+ */
+void UBoat::align_components()
```
frozenblit, line 1346: This removes an item listed twice. It misses a component saved only as a `Traveller`, such as a loose plank on the deck. That item still takes the first slot of its objtype and pushes the boat's own component out. The next save then keeps it as a `Component`.

What if `readProperties()` read the `Component` lines first, and the `Traveller` branch skipped anything already in `Components`? Then the "keeps its last place" rule could go.

pol-core/doc/core-changes.txt
```diff
@@ -57,6 +57,13 @@
            `Equip Sandals 1940, 1965, 1109, 1150-1154` picks one of the four entries with equal
            odds, then a value inside it when that entry is a range. EquipFromTemplate() makes the
            pick per item on every call, and skips and logs a color it cannot read.
+09-14-2026 Ins:
```
frozenblit: This needs to go at the top of `POL100.3.0`, dated near the merge, then re-run `buildcorechanges.py`. Also, admins call these "components", not "parts".

pol-core/pol/testing/testboat.cpp
```diff
@@ -0,0 +1,366 @@
...
+  auto* realm = Core::gamestate.Realms[0];
+  const Core::Pos4d pos( 10, 50, -4, realm );
+
+  // the tillerman listed twice
+  {
+    auto p = parts_at( pos );
+    auto* boat = load_boat(
+        pos, { component( p.tillerman ), component( p.tillerman ), component( p.portplank ),
+               component( p.starboardplank ), component( p.hold ) } );
+    if ( !boat || !step( boat ) )
```
frozenblit: On a failed check, this leaves the boat in the world for the tests that run after it. Could you destroy it before returning?

frozenblit (commented):
Looks right to me. Every loop over `Components` skips a null slot. The one change I'd ask for is the `Traveller` point on `align_components()`.

murderinc:
@frozenblit Rebased onto master; your three points are done. Your inline comments will show as outdated after the force-push — replies are in the threads.

Four calls I'd rather you made:

1. **Should a component-objtype `Traveller` stay a traveller?** It no longer takes the boat's own component's slot — but if that component was destroyed before the save, it fills the empty slot and `reread_components()` makes it the boat's plank, permanently. Untested: I didn't want to pin behaviour you might change.
2. **Is a slotless component meant to be stranded?** Nothing past the last slot is moved or turned again, and those items are `movable(false)` with decay off, so one becomes permanent open-water scenery. Before, it was index-paired into some slot and at least travelled.
3. **Move the de-duplication out of `align_components()`?** It can't go — drop it and four checks fail, because two `Component` lines can name one item, unlike the Traveller/Component pair. It could move into `readProperties()` as erase-and-append. Either suits me.
4. **Keep the `Traveller`-side `std::find` guard?** Correct, but the de-duplication already absorbs the case, so no test can tell it's there.

Separately, on its own branch: `create_components()` has the same slot-shift on the create path.

Also unrelated: `support/boats.cfg` gives multiids `0x2D` and `0x2F` 26 `Hull` lines where `0x2C` and `0x2E` have 27 — the only asymmetric facing set in the file, and no duplicate coordinates, so it looks like a dropped line. A ship of that family loses a hull piece at two of its four headings.

pol-core/pol/multi/boat.cpp
```diff
@@ -1286,8 +1287,9 @@ bool UBoat::turn( RELATIVE_DIR dir )
 
 bool UBoat::is_component( const UObject* obj ) const
 {
-  return std::any_of( Components.begin(), Components.end(),
-                      [obj]( const Component& c ) { return c.get() == obj; } );
+  // An unfilled slot holds a null, so the null has to be ruled out before the comparison.
+  return std::any_of( Components.begin(), Components.end(), [obj]( const Component& c )
+                      { return c.get() != nullptr && c.get() == obj; } );
```
frozenblit: wouldn't it be enough to check for obj != nullptr?

murderinc: Yes, it's the same condition. Comparing a non-null `obj` with an empty slot is already false, so only a null `obj` needs ruling out. `is_component()` now returns false for a null `obj` before the loop, and the lambda is back to master's. No caller passes a null today. The missing-component case in `boat_load_test` pins it: without the guard, that save's empty slot matches a null. Done in 7fd54f647.

frozenblit (approved):

PR #967, opened 2026-09-14
Keep selling past an item that has left the backpack

murderinc:
oldSellHandler returned as soon as a serial in the sell answer was not at the top level of the seller's backpack. By then the items listed before it had already been moved into the vendor's bought container, but the gold is only created after the loop, so they were kept without being paid for. The rest of the answer was never looked at, no EVID_MERCHANT_BOUGHT reached the vendor, and the sell window stayed armed. A player gets there by putting a listed item in a bag, or using it up, before accepting the window.

Skip that item instead, as the buy handler already does for a ware it cannot find. The scripted handler is left alone: it moves nothing itself, so dropping the whole answer there loses nothing.


pol-core/pol/module/uomod2.cpp
AsYlum-, line 936: This will probably need the same treatment as in oldSaleHandler because we also have `ScriptedMerchantHandlers` in serverspecopt.cfg

PR #968, opened 2026-09-16
Fix clean.py: stop deleting files tracked by git

### LINKED ISSUE #465: clean.py in rootDir deletes .cmake files in rootDir/cmake

AsYlum-:
findCmake treated every `*.cmake` as generated, so a run removed all 30 hand written modules under cmake/, all nine pol-core/*/CMakeSources.cmake and around 400 vendored files under lib/, among them FindANTLR.cmake and source Makefiles that a rebuild does not bring back. It now matches the generated names exactly and does not descend into .git, lib or testsuite. A configured build folder is removed whole, unless git tracks something in it as it does for bin-build.

Separately, every candidate list is filtered against git ls-files, so nothing in the index can be removed even if the classification goes wrong again.

Three neighbouring faults went with it: findBinaries looked in pol-core/bin, which no longer exists, and raised FileNotFoundError right after the deletion; findBoost never matched boost_1_91_0-1; findEmpty walked .git and would have removed .git/refs/tags. Added -n/--dry-run to list what a run would delete.

Fixes #465


clean.py
```diff
@@ -7,20 +7,78 @@
...
-import shutil
 import re
+import shutil
+import subprocess
+
+
+# Files CMake generates beside the sources or inside a build directory, matched by
+# exact name on purpose: the tree also holds hand-written cmake modules (cmake/,
+# pol-core/*/CMakeSources.cmake) that a ".cmake" suffix match would destroy.
+GENERATED_NAMES = frozenset((
```
frozenblit, line 18: is there really no difference between generators or windows/linux?

frozenblit (approved):

clean.py
AsYlum-: Seems to work, ran it on debian 12 vps and dryrun lists files then deletion also seems to be working fine, no problems in git status afterwards. All looks clean

PR #969, opened 2026-09-16
Allow color ranges and color lists in equip.cfg

### LINKED ISSUE #567: equip.cfg & EquipFromTemplate
So in equip.cfg we can only assign a single color to an item, I'd like to request that this be expanded in two ways:

Allow for a range to be added with - being the delimiter between the two high and low values.
When an item is assigned a color from EquipFromTemplate function, if it is a range value, the function would select a value from the range, so 100-200 assigned to an item in equip.cfg could be colored any value from 100 up to and including 200.

Allow for an array to be added whereby individual colors can be added, these would be broken up by the comma character.
If an array of colors are added, each entry can either be a single color value or a range of colors. The EquipFromTemplate function would read in the array and randomly select one of the entries, if the object selected is a non-range value, the color is assigned to the item as normal, if it is a range value, it would be piped off to select a value within the range.
```
Equipment Healer
{
    Equip            Robe           1701-1754
    Equip            Sandals        1940, 1965, 1109, 1150-1154
    Equip            QuarterStaff
    Equip            LongHair       0x461
}
```
This is a silly example of what an entry could look like, notice LongHair has a single value which is how it works now, Robe has a Ranged value assigned to the color slot and Sandals has an array value assigned to the color slot with the last entry in the array being a Ranged entry.

AsYlum-:
The color slot of an Equip/Weapon/Armor line took a single value, so an equipment template dressed every character it touched identically. It now also accepts an inclusive range written low-high, and a comma separated list whose entries are single colors or ranges:

    Equip Robe      1701-1754
    Equip Sandals   1940, 1965, 1109, 1150-1154

EquipFromTemplate() picks a list entry with equal odds per entry, then a value inside it when that entry is a range, once per item on every call.

The slot is now read as the whole rest of the line rather than one whitespace token. strtoul still takes its base from the prefix and still stops at trailing text, so decimal, hex and text after a value behave as they did. An entry that cannot be read is dropped and reported, and the item is created anyway, uncolored if nothing in the slot parsed.

Resolves #567


pol-core/pol/testing/testequipcolors.cpp
```diff
@@ -0,0 +1,81 @@
...
+// as the grammar: a reversed range is kept (the pick swaps it), text after a value is ignored the
+// way the one-colour form always ignored it, and one bad entry costs only itself.
+void equip_colors_test()
+{
+  // no colour at all
+  check( "", "" );
+  check( "   ", "" );
+
+  // single value, decimal or hex, as before
+  check( "1701", "0x6a5-0x6a5" );
```
frozenblit: shouldn't you also have a check with an actual single value?

frozenblit (dismissed):

pol-core/pol/testing/testequipcolors.cpp
AsYlum-: Makes sense. Should be fixed now.

frozenblit (approved):

PR #976, opened 2026-09-20
Sell only what the sell window offered, and pay for all of it

murderinc:
Supersedes #967, which fixed the first entry below, and only in the core's handler.

A 0x9F sell answer was taken on trust. Both handlers sold whatever serials it named, read as many entries as its count claimed, and told the vendor about a sale the seller was not always paid for. `SendSellWindow()` now keeps the serials it listed and the vendor it listed them for in `ClientGameData::vendor_sell_offer`, and both handlers sell against that list.

- `answered_sell_window()` replaces the vendor checks in both handlers. When a window for another vendor has since taken `gd->vendor`, it drops only `vendor_sell_offer` and leaves those fields to that window.
- `sell_answer_entries()` reads no more entries than `msglen` leaves room for, whatever `num_items` claims. `sellhandler()` ignores an answer too short to carry a vendor serial and a count.
- `item_for_sale()` skips an entry the window did not list, a serial named twice, an amount of 0, a container whose `count()` is not 0, and an item `find_toplevel()` no longer finds.
- `oldSellHandler()` used to return at the first item that had left the top level of the pack, with the items before it already moved to the vendor and unpaid. `sellhandler()` returned there too, before sending its event.
- `hand_over()` moves the sold item with `Items::move_into()`, which gives it a free slot in the vendor's container. What is left of a split stack goes to its own pack slot, then any free slot, then the seller's feet.
- `pay_seller()` pays in stacks of at most 60000 and puts a stack the backpack refuses at the seller's feet. `_create_item_in_container()` gained a `bool* taken`, so gold a pack script consumes is not paid twice.
- `oldSellHandler()` keeps the running total in a `uint64_t` and does not sell an item that would take it past `INT_MAX`. `SYSEVENT_MERCHANT_BOUGHT.amount` is what `pay_seller()` returned rather than what the sale was worth.
- `send_vendorsell()` returns the serials it listed. It applies `VENDOR_BUYABLE_CONTAINER_FILTER` before measuring an entry against the packet, so an item the filter drops cannot keep the list from going out.
- `mf_SendSellWindow()` returns `BError( "The sell list is too long to send" )` instead of 1, and calls `send_clear_vendorwindow()` first. The merchant's containers go out as worn items before the list does.
- `oldBuyHandler()` and `buyhandler()` reset `vendor_sell_offer` along with `gd->vendor`, so a buy window no longer arms the sell handler.
- `oldBuyHandler()` moves a purchase with `Items::move_into()` instead of relocating it into `tobuy->slot_index()`.
- `_create_item_in_container()` relocates into the slot `can_add_to_slot()` settled on. It passed `item->slot_index()`, which is 0 for an item that has never been in a container.
- `getitem.cpp` falls back to a free slot for what is left of a partly lifted stack, the same fallback `hand_over()` uses.
- `SuspiciousActs::SellAnswerWithoutWindow()`, `SellAnswerNotListed()` and `SellAnswerOverflows()` log under `pol.cfg ShowWarningItem`. An answer with no entries is how a client closes the window, and is not logged.

**Behaviour change**

- `SendSellWindow()` can return an error.
- A sell answer to a buy window sells nothing. A script that bought from players by opening a buy window has to open a sell window.
- With `ScriptedMerchantHandlers=1`, the shopping list leaves out what the core would not sell.
- `SYSEVENT_MERCHANT_BOUGHT.amount` is what the seller was paid.
- With `UseContainerSlots=1`, created items, purchases and sold items take a free slot instead of slot 0.

`core-changes.txt` carries the full list.

**Testing**

- 37 shard cases in `test_client_vendor_sell.src`, `test_client_vendor.src` and the new `test_client_container_slots.src`. Each was seen failing without its own fix.
- New fixtures: `ai_vendor.src` passes the merchant events on to the test, a 0x9E send hook appends an entry the core never listed, and `pack_eats_gold.src` is a CanInsert script that destroys gold.
- The test client reports the container slot the core gave each item, and `ResourceManager` gained `AddCleanup`.
- Not covered: the 0x3B that closes the merchant window. The test client has no parser for it.
- Not covered: a sell window followed by a buy window for the same vendor. This PR does not change it.
- Not covered: the refusal branches in `buyhandler()` and the trade window. The `can_add()` check above each makes them unreachable without a CanInsert script.

A follow-up branch gives items in a secure trade window and the copies a corpse takes a slot of their own. It builds on this one.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


murderinc:
@AsYlum- took a while, but now it should be completely covered, and then on top of this PR will be another.

docs/docs.polserver.com/pol100/corechanges.xml
```diff
@@ -2,9 +2,56 @@
...
 	<header>
 		<topic>Latest Core Changes</topic>
-		<datemodified>09-20-2026</datemodified>
+		<datemodified>09-21-2026</datemodified>
 	</header>
 	<version name="POL100.3.0">
+		<entry>
+			<date>09-20-2026</date>
+			<author>Ins:</author>
+			<change type="Fixed">a sell answer naming an item that was no longer at the top of the seller's backpack -<br/>
```
frozenblit, line 11: Can you let Claude shorten the core-changes according to the new CLAUDE.md? The default prose is very verbose and makes it hard to understand what actually changed

murderinc: Done on both #976 and #966.

frozenblit (approved):

